Issue #16, opened 2016-12-21
As an admin I should be able to add clients, volunteers, donors to the database without creating user accounts for each.


jasonsreynolds:
Currently the only way to add new users to the database (aside from running the "createAdminUser.js" script) is to log out of the app and go through the "sign up" process. The admin user should have a way to add new users to the database from the admin panel without logging out.

nicholashu:
I might give this a go. Any tips?

DealPete:
Right now you can edit a client from the "Client Database" tab by clicking on the edit button in his row. The edit screen allows you to change any details about the client. I think the easiest way to implement this would be to have an "add new client" button at the bottom of the table that automatically adds a new client to the database with some default settings, then takes the admin user to the edit page to fill in the details.

nicholashu:
Should the newly created volunteers, clients and donors have user accounts or just database entries?

DealPete:
Just database entries.

DealPete:
Hi Nicholas,

I just realized that it's not going to work to just create database entries in customer, donor, and volunteer. Currently these get created dynamically from the corresponding user accounts, with the id set to the user's id. Adding an entry directly to one of these collections could result in an id conflict. The best solution to this may be to create a dummy user account, but I'll leave it to you to decide.

nicholashu:
I found that out, I set up a flag on the volunteer creation that only deletes and sets the ._id if manualAdd is not true. Otherwise it creates an ObjectID (e.g., 587c2fcade02ac8be4ec8779). And will leave the ID's starting with 10000 to just users. Could either create a different ID set for manual adds e.g., m1000.

khawajah:
Hi..can i take this issue??

khawajah:
can i have more details for this issue please!!
i figured that you need is to let the user admin to add new user without logout ... so add new button called (add new client) with some setting after that should take the user admin to edit page  to fill in the details... is my explain is right? and should i let the user admin when he create new user to choose  (volunteers, clients or donors )

nicholashu:
i've almost finished this. Just need to work out how to make a dummy user account for creation.

khawajah:
you need to create dummy user in database?? can u give me more details please!!
!

DealPete:
Hi Khawajah,

Nicholas has been working on this one for awhile. If you'd like to help out, feel free to check out some of our other issues.

nicholashu:
[...]

PR #32, opened 2017-01-17
As an admin I should be able to add clients, volunteers, donors to the database without creating user accounts for each. #16

nicholashu:
fixed staging conflicts, added feature.


nicholashu:
Hold up, there seems to be an issue with the edit page now.

nicholashu:
issues with merging fixed

nicholashu:
I have an issue where if a new volunteer is created manually, it can't use the $update function in the controller as it is trying to do "PUT /admin/volunteers" instead of  /admin/volunteers/:volunteerId any ideas?

DealPete:
That error suggests that the "_id" is blank for that user. Are you using a different id field for the manually added users?

nicholashu:
It seems my objectId method of _id is flawed. I will have to create dummy users for manual additions.

PR #35, opened 2017-01-18
Expose title property to views in order to fix blank title in HTML

### LINKED ISSUE #33: Title is blank in the HTML
In the file app/views/layout.server.view.html line #5 the {{title}} variable is not set when the template is rendered.

kenjiO:
One of the views attempts to use {{title}} which is not set.  This fixes it by setting res.locals.title from the config.app.title property so that views have access to the application title.  Also added a title property to the production env properties.
Closes #33

### DIFF
--- a/config/env/production.js
+++ b/config/env/production.js
@@ -1,5 +1,8 @@
 'use strict';
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


MAF27:
Thanks kenjiO, looks good, I'll merge it.

PR #36, opened 2017-01-19
added session time out function

Rusticman:
In order to track the activity of the user, I created a function in the "authentication.client.controller.js" file.  I used the following stackoverflow article as a guide:

http://stackoverflow.com/questions/667555/detecting-idle-time-in-javascript-elegantly

I used the second solution as it uses  pure javascript. The function works by firing once the user has successfully logged in or signed up. It uses an Angular service - ```$timeout``` which operates exactly like ```setTimeout```. After the time period is up, the ```$timeout``` function finally executes.  It checks if the ```flag``` variable is ```true```. If so, it means the user has fired ```keydown``` or ```mousemove``` which executes the ```$timeout``` function again and sets the ```flag``` variable to ```false``` again. If the ```flag``` variable is false when the ```$timeout``` function executes, a ```$http``` request is sent to the server to sign out. I then used the ```$window``` service to force the URL to the 'signin' URL. I tried to use the ```$state``` service but, I must admit, I'm at a loss. I'm still figuring out Angular. 

On lines 21 & 41, you will see the ```$timeout``` functions. I've put the times at 5000 (5 seconds) in order for you to test. Please feel free to contact if you need any further clarification.

Some of the changes below aren't really changes, as I have used a lot of ```console.logs``` to figure out how the app works. Good job on the site by the way, very impressive.

### DIFF
--- a/admin-config.json
+++ b/admin-config.json
@@ -1,7 +1,7 @@
 {
-	"username": "rusticman",
-	"password": "foodbank90",
-	"email": "russellwheatley85@gmail.com",
-	"lastName": "Wheatley",
-	"firstName": "Russell"
+	"username": "admin",
+	"password": "password",
+	"email": "john@email-provider.com",
+	"lastName": "Admin",
+	"firstName": "Johnny"
 }
--- a/admin-config.json
+++ b/admin-config.json
@@ -1,7 +1,7 @@
 {
-	"username": "admin",
-	"password": "password",
-	"email": "john@email-provider.com",
-	"lastName": "Admin",
-	"firstName": "Johnny"
+	"username": "rusticman",
+	"password": "foodbank90",
+	"email": "russellwheatley85@gmail.com",
+	"lastName": "Wheatley",
+	"firstName": "Russell"
 }
--- a/app/controllers/users/users.authentication.server.controller.js
+++ b/app/controllers/users/users.authentication.server.controller.js
@@ -19,12 +19,12 @@ exports.signup = function(req, res) {
 	// Init Variables
 	var user = new User(req.body);
 	var message = null;
-	
+
 	// Add missing user fields
 	user.provider = 'local';
 	user.displayName = user.firstName + ' ' + user.lastName;
 
-	// Then save the user 
+	// Then save the user
 	user.save(function(err) {
 		if (err) {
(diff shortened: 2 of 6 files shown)


DealPete:
Hi Rusticman,

The code that you added looks good. I looked into why $state.go wasn't working, and the problem is that the client still thinks that the user is logged in. When AuthenticationController queries the Authentication service to find out the user (on line 50), it redirects the user to 'root'. $window works properly because going to "/" requests the whole application again from the server. Ideally the Authentication service could be informed that the user has logged out.

Rusticman:
I've changed the code so that ```Authentication.user``` is set to ```null``` after the user has been logged out server-side. I've removed the ```$window``` service and used ```$state.go``` to push it to the 'route.signin'.  Please advise if you require any further changes. Sorry about the ugly presentation of the code, it doesn't look like that on my text editor!

DealPete:
Hey, I tested that last change and it works. Go ahead and get this ready to merge with the codebase and I'll accept the request.

Rusticman:
Sorry, not sure if I understand - do you mean make a pull request to the master branch?

DealPete:
Sorry, I should have been more clear: I can't accept the pull request like it is. A pull request should be in a state to be merged into the parent repository (to freecodecamp/food-bank-app/staging) without needing changes from the maintainers. You'll have to take out the console.log statements, change the timeout from 5 seconds to 15 minutes, and so on.
I noticed that you changed the admin-config.json file as well. It's meant to be changed, but you shouldn't submit your changes in a pull request :)

Rusticman:
I think I've caught everything required for the merge, I went through every commit I've made in order to remove any non-necessary changes. I've also changed the timeout functions to 900000 which equates to 15 mins. Let me know if there are any more that slipped through the net :)

Rusticman:
Instantly found one I didn't catch, the lat/long in the driver controller. I'll get on it.

DealPete:
Looks good Rusticman, thank you for your hard work. I just merged this manually so I'm going to close the pull request now.

Rusticman:
Thanks DealPete. That was my first pull request so I'm pretty happy about that. I learnt that it's probably a wise idea to open a new branch so that I don't push every commit to the pull request 👍

PR #38, opened 2017-01-27
Fix client admin authorization code

### LINKED ISSUE #34: When I am not logged in as an admin, I can still navigate to some admin routes instead of getting redirect to the login patge
/#!/admin/volunteers
/#!/admin/donors
/#!/admin/questionnaires

kenjiO:
Move code that checks for an admin user on the client out of the controllers.  Accomplished by

- Changing the Authorization factory to a provider so it can be used in the route config areas.
- Adding a method to Authorization to check for an admin user
- Adding the resolve property to the ui-router routes that require admin access
- Setting the resolve property to the Authorization method to check for administrator

closes #34

### DIFF
--- a/public/modules/customer/config/admin-customer.client.routes.js
+++ b/public/modules/customer/config/admin-customer.client.routes.js
@@ -1,8 +1,8 @@
 'use strict';
 
 // Setting up route
-angular.module('customer').config(['$stateProvider',
-	function($stateProvider){
+angular.module('customer').config(['$stateProvider', 'AuthenticationProvider',
+	function($stateProvider, AuthenticationProvider){
 		// Customer state routing for admin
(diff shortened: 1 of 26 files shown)


kenjiO:
I can also add similar logic to check that a user is logged on for all the user routes requiring a user to be logged in.  Let me know if you want me to work on that too.

DealPete:
looks good to me. It would be great if you updated the user routes too. btw, what's the purpose of the $timeout(... ,100) before redirection?

kenjiO:
The example I found online that I based this code on had the timeout in it but didn't explain it.  I tried removing it but it didn't work without it so I left it in.  The example had a timeout of 1000 but I changed it to 100 so there is less delay.

PR #39, opened 2017-01-30
Feature/change map

Rusticman:
a few things to note:

1. I gave the map a fixed height in order for it to display.
2. The linter alerts to the 'google' global object not being defined which I believe it is checking from within the scope of the controller.  It works fine because it's global. 
3. I've taken out my google API and used the ```{{googleKey}}``` in the 'app/views/layout.server.view.html' file. 

I did squash the twelve commits since the merge on Jan 24th, but then I made another change which I'm having difficulty squashing into a squashed commit so apologies for the two commits.  If there is anything you want me to change/do again, please let me know. 

Cheers,

Russ


DealPete:
Hi Rusticman,

When I run your patch I get 66 linting errors, and when I go to "route assignment" I just see a blank white box where the Google Map should be.

Rusticman:
Apologies, Pete. It was an unresolved merge issue. I've removed the problems but I'm not happy with it because I wanted a cleanly squashed commit for you. Give me some time to figure out how to use mergetool, and I'll pass along a squashed commit.

DealPete:
Good work, that's working. I can't accept the pull request, though, until you fix the linting errors. The reason it flags "google" as undefined is that we don't use global variables in the app (besides objects like document and window). I presume the google object comes from angular-google-maps... in that case it should expose a dependency you can inject into the controller.

Rusticman:
Ok, I'm looking at a couple of solutions to this problem. I'll let you know when I have something working

PR #40, opened 2017-02-01
Feature/change map

Rusticman:
Hi Pete, I started from scratch because I was tying myself in knots trying to squash commits and resolve merges. The commit, "changed to google map API" has every change I've made for this particular issue. 

I tried to find a solution that involved the google maps object and the marker clusterer object not showing up in the linter.  

Initially, I thought it might be possible to make an API call to receive a static version of the library. However, according to this SO question, it isn't possible:

http://stackoverflow.com/questions/28490425/how-to-use-google-maps-api-v3-without-add-script-tag-on-html

A script tag, as far as I've read, seems necessary to create the global variables.  I followed your advice in the previous pull request and created a service for these two particular variables and injected them into the controller.  The service is in this sub-directory:

"public/modules/driver/services/driver.client.service.js"

If there is another solution that I haven't followed up then I'd be happy to implement it. Let me know what you think.


DealPete:
It must be possible, because the angular-google-maps plugin is just a wrapper around the google maps API.

DealPete:
I just had a look at the source for angularjs-google-maps, and they use the $window service to get global variables. i.e. `$window.google`

Rusticman:
That's great! I was about to start looking under the hood at ```angular-google-maps``` to figure out what magic they're using that it didn't require a script.  I'll get on it now.

Rusticman:
I used the window service. Would you like me to squash the last two commits together?

DealPete:
Yes, please. You should always squash your commits.

Rusticman:
Let me know if it's ok.

DealPete:
It's working. Congratulations, Rusticman.

Rusticman:
Thanks, Pete. I was wondering if there was an issue that was more urgent than this one:

https://github.com/freeCodeCamp/food-bank-app/issues/10

I'm happy to tackle this one, but maybe there's another one that you feel is more urgent.

DealPete:
The issues labelled "help wanted" are slightly more urgent, but any issue you solve would be a valuable addition to the app.

PR #42, opened 2017-02-09
integrated driver's route

Rusticman:
layout.server.view.html ---> I added the geometry library in order to measure which client is furthest away from driver.

routes-driver.client.controller.js ---> I noticed a comma that shouldn't be there.

user-driver.client.controller.js  ---->  HTML 5 is used in order to ask the driver's geolocation, from that point, the controller works out the furthest client and makes that the destination.  It then figures out the optimal route for the driver to take.  I've left notes for you to follow along with the code.

user-driver.client.view.html ----> I've integrated the map alongside the list of clients so it matches the admin driver route layout.

Design issues for you to decide upon:
1. Two images as markers; 1 for the client and 1 for the driver. Your decision on suitability.
2. Is a marker clusterer appropriate in this instance?

Error handling:
1. If browser doesn't support HTML5 or if the driver chooses not to share his location, then a window  pops open with an error message.  The driver still has a route, but it starts from the geoToronto lat/lng variable. This could possibly be the lat/lng of the company.

2. For whatever reason the route doesn't work, the driver gets an error message. It didn't work when I set it to Liverpool and Toronto for obvious reasons, but it should be good for the local area.


DealPete:
Hi Rusticman,

I get an "Error The geolocation service has failed" error whenever I go to the page. In Safari, it refuses to use the API over an insecure connection. In Google and Firefox I get this error, then it shows me my location, but doesn't put on a marker, even for the customer living two doors down the street.

PR #43, opened 2017-02-10
User ui-router resolve to check for logged in user

kenjiO:
This is applying the same logic from what I did in  [0a3832e](https://github.com/freeCodeCamp/food-bank-app/commit/0a3832e020f277336a7c5c85481c528c126a2df9) to the user routes.  

I created a `requireLoggedIn` method in the `AuthenticationProvider` that will redirect to the login page if there is not a logged in user.  Then I added this method to the ui-router resolve property of every route that requires a non-admin to be logged in. 

I think that this could be further improved to check for a role rather than simply if someone was just logged on.  This could be done by making a similar method to `AuthenticationProvider.requireAdminUser` for each type of role(i.e. requireCustomerUser, requireDonorUser).  I wanted to figure out how to just have one method requireRole(role) that you can use for any given role but I couldn't find a way to do that.

### DIFF
--- a/public/modules/customer/config/user-customer.client.routes.js
+++ b/public/modules/customer/config/user-customer.client.routes.js
@@ -1,8 +1,8 @@
 'use strict';
 
 // Setting up route
-angular.module('customer').config(['$stateProvider',
-	function($stateProvider){
+angular.module('customer').config(['$stateProvider', 'AuthenticationProvider',
+	function($stateProvider, AuthenticationProvider){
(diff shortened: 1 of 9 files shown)


DealPete:
This looks good. You suggestion is sensible, but I'm still somewhat confused about the difference between roles and accountTypes. It would be very helpful if you could deal with Issue 37 (https://github.com/freeCodeCamp/food-bank-app/issues/37) first.

PR #47, opened 2017-02-13
reinstated food service

Rusticman:
The food service had been accidentally removed in this commit:

https://github.com/freeCodeCamp/food-bank-app/commit/cfbda6be313f15a0b8160569742e40187eb6a783

I should have the driver route directions ready for a new commit & approval at some point today hopefully. 

Also - After updating the staging branch with a pull request, there are 6 linting errors on the CLI, and in the browser console there's a warning not to use XMLHttpRequest because it is synchronous. I'll include a factory function in my next commit which will make the request async and also remove the linting errors.


DealPete:
Thank you for pointing this out. We have fixed the food issue, and I'm going to fix the linting errors now.

PR #48, opened 2017-02-13
driver route integration

Rusticman:
Hi Pete, 
             Sorry I had to create another pull request, I'm hitting a few errors trying to squash commits and permissions are being denied. 

layout.server.view.html ---> I added the geometry library in order to measure which client is furthest away from driver.

routes.driver.client.controller.js ----> The previous commit added a function to get the geolocation of the city.  I refactored it to a service so that it was reusable in another controller. This was made tremendously easy because of yzhbankov previous work as I basically followed his code. Thank you, yzhbankov.

I also noticed that markers were rendering twice, so I fixed that problem by removing duplicates of the markers and the marker clusterer.

user-driver.client.controller.js ----> I've changed this in order to catch the various problems that can be encountered whilst running the geolocation method.  The flow is as follows:

1. user's location captured, clients mapped out with optimal route. The driver's location being the original point and the client furthest away being the destination.
2. If geolocation fails because the browser doesn't support it, the user didn't give permission, or it silently fails because no permission was given/removed. The route is mapped with the driver's position being the geolocation of the city. This could potentially be the address of the foodbank although I couldn't see the address in the data object passed by the server when I hit this route 'api/settings'. Or maybe even the volunteer's address, not sure if that breaks any privacy rules though.
3. If route fails, first change the driver's position to the city geolocation and try mapping the route again.  If route fails again, a function runs that renders the markers without any route.

I've checked this in a few of the browsers and even with addresses in different countries, the markers render, so you should see you neighbour's house this time :)

driver.client.services.js ----> This is the service I mentioned previously.

user-driver.client.view.html ----> the new view to accommodate the map alongside the list of clients.

If there are any other problems, please don't hesitate to let me know.


DealPete:
Hi Rusticman,

I'm getting inconsistent results with different browsers. In Safari it flat-out refuses to do geolocation over an insecure connection, and throws an error. In Chrome, it uses the location from App Settings as my location (perhaps because geolocation is failing). In Firefox it does put the little blue circle at my actual address, but it doesn't map a route from there to any of the customer addresses. This could be because there's an address overseas... why is the most distant client used as the destination?

I'm going to put up an issue asking for someone to implement TLS (https), because it seems like we may need it for the issue.

Rusticman:
I resigned myself to the fact that safari wouldn't work as the callback never comes as noted in these articles:
http://stackoverflow.com/questions/3686630/geolocation-not-workin-on-safari-5-x-on-windows-7-xp
http://stackoverflow.com/questions/3791442/geolocation-in-safari-5
http://stackoverflow.com/questions/32948143/html5-geolocation-not-working-on-safari-browser

It seems to be more of an issue for windows operated machines which mine is.  The chrome browser hasn't failed to locate my machine so I don't know what that could be. Do you have more than 23 waypoints? That's the maximum quota for the route direction service.  If you have a client overseas then the route service will not work, another function will kick in that displays the markers without the route.

The reason why the most distant client is used is because the service needs a starting point and a destination. Unfortunately, the direction service doesn't calculate the most optimal route by just passing it a bunch of waypoints.

Rusticman:
did you have an overseas client when you used the service in google chrome? The flow is
1. geolocate  driver and create route. 
2.If this fails because of overseas client, it changes the driver's position to the city. It does this because it requires a starting point.

DealPete:
If Google allows you to calculate routes, it should be possible to calculate each route then check which is the shortest. Having the longest route set by default doesn't make any sense - imagine how much longer your route will be if you always travel to the most distant target!

Rusticman:
You're right, the route won't always be optimal, although I'm not sure it will always be the longest.  I guess I was thinking linearly which isn't smart considering the clients will more likely be scattered. It's a much better idea to loop through the possible starting point/destinations for the shortest route.  I'll work on a function that implements this solution.

To be honest, I've been second guessing what you want which is my bad because I should be asking questions about what it is you want. 

1. If the route fails, are you happy to try again by changing the starting point (i.e food bank address or driver's address)? or would you prefer to go straight to mapping the clients with markers without a route?

2.If geolocation fails (not available, errors or simply doesn't call back) would you be happy to change the starting point? or map the clients without a route?

3. I am still interested in the errors you are receiving, it's a problem if you're not able to get a route through the chrome browser.  Are you able to get a route if the clients are in your area?

DealPete:
Hi Rusticman,

I reread your description, and I think I misunderstood what you had implemented. Do you mean that the algorithm maps out a route stopping at each of the waypoints, with the start of the route being the volunteer's address, and the end of the route being the furtherest client? If that's the case it should be fine. We could improve on that by using some sort of travelling salesman algorithm, but that's an issue for another time.

I think it should be fine to use the volunteer's address as the starting point when geolocation fails. I would implement it thus:

1. First try geolocation to get the volunteer's current position.
2. If that fails, use the volunteer's home address.
3. If that fails, fallback to displaying markers.

Ideally, if one of the waypoints is unreachable (because it is overseas, say, or the address can't be found), it should exclude that waypoint and calculate a route based on the others.

Rusticman:
> Do you mean that the algorithm maps out a route stopping at each of the waypoints, with the start of the route being the volunteer's address, and the end of the route being the furtherest client?

It tries geolocation instead of volunteer's address first, but yes, this is it :)

> I think it should be fine to use the volunteer's address as the starting point when geolocation fails. I would implement it thus:

> First try geolocation to get the volunteer's current position.
> If that fails, use the volunteer's home address.
> If that fails, fallback to displaying markers.

Perfect, I'll implement this.

>Ideally, if one of the waypoints is unreachable (because it is overseas, say, or the address can't be found), it should exclude that waypoint and calculate a route based on the others.

If the ```directionService``` passes the erroneous waypoint/waypoints in the error function, then this shouldn't be a problem.

PR #51, opened 2017-02-15
revised map

Rusticman:
As per what we discussed in the last pull request. There is no capability to find out which waypoint is breaking the directions unfortunately.  Let me know if there are any further changes to be made.


DealPete:
Hey Rusticman,

It's coming along, but I'm still seeing some issues.

1. It takes several seconds for the list of deliveries to appear. The list should appear as soon as possible and not be dependent on the google maps API call.
2. If my home address is invalid I get an "Error: undefined is not an object (evaluating 'response.data.results[0].geometry')".
3. When the page loads, it centres the page on my street address (107 Springdale Street, St. John's, Newfoundland, Canada), but displays markers instead of a route, even though all four of the waypoints are legitimate addresses (I verified this with Google Maps):

101 Springdale Street
103 Springdale Street
25 Empire Avenue
21 Harvey Road

Rusticman:
Hi Pete,
            I've had a look at the problems you've encountered:

>It takes several seconds for the list of deliveries to appear. The list should appear as soon as possible and not be dependent on the google maps API call.

Good catch, it shouldn't wait for the map and I've changed that now.

>If my home address is invalid I get an "Error: undefined is not an object (evaluating 'response.data.results[0].geometry')".

This is from the Geolocation service I created (not the HTML5 geolocation). The geolocation service is used to find the driver's address and the city location. I have a hunch that the error is from finding the driver's address as this is necessary for creating a route if HTML5 geolocation fails.  I've created an if statement should the response === 'NO_RESULTS" to stop the code breaking.

>When the page loads, it centres the page on my street address (107 Springdale Street, St. John's, Newfoundland, Canada), but displays markers instead of a route, even though all four of the waypoints are legitimate addresses (I verified this with Google Maps):
101 Springdale Street
103 Springdale Street
25 Empire Avenue
21 Harvey Road

I recreated this. I used your address as the driver's address, and the other four as client's addresses.  I was able to generate a route using Opera, Chrome, Microsoft Edge, Firefox and Safari. My hunch is that issue 2 & 3 might be related.  I think the response to the driver's address === 'NO_RESULTS'.

I've added two console.logs in the geolocation service for you to test, they look like this:
```console.log('geolocation driver address response is:', response.data.status);```
```console.log('geolocation city response is:', response.data.status);```

Could you run the latest commit and let me know, please? Thank you.

DealPete:
Issue 1 and 2 are fixed, but now, for every browser I try, I get a blank grey field where the map should be (after it does some loading). As well:

"geolocation city response is: OK"
"geolocation driver address response is: OK"
"geolocation driver address response is: OK"

Is the latter message supposed to appear twice?

Rusticman:
> Is the latter message supposed to appear twice?

No, it shouldn't be rendering twice, that could be the problem. I'll try and replicate that so I can fix it.

Rusticman:
I can't recreate the grey map. I have stopped the geolocation service from firing twice. Could you let me know what you get.

 I'm wondering if the grey map is from the map intialization actually. I changed the code slightly from this:
`     self.mapObject = new googleObject.maps.Map(document.querySelector(".googleMap"), {
 	           center: self.foodBankGeo,
 	           zoom: 12
 	         });`

to this:
`self.mapObject = new googleObject.maps.Map(document.querySelector(".googleMap"), {
						 center: new googleObject.maps.LatLng(0, 0),
	           zoom: 1
	         });`

This essentially intialises the map so you see the entire map before the route/markers render. I thought it looked better than zooming into the city location initially.

Rusticman:
From what I've read, it is a distinct possibility that it could be the change to `center` or `zoom` in the options object passed to the map that makes it fail silently (go grey). 
http://stackoverflow.com/questions/21210533/google-maps-grey-map
http://stackoverflow.com/questions/11749742/google-maps-api-v3-gray-box-no-map

another common problem is the CSS settings. I think it is more likely to be the above.

DealPete:
I was able to get it working after with a local database. I'm going to accept this for now, and if I can't get it to work with my remote database put it up as an issue.

DealPete:
good work, btw 👍

PR #54, opened 2017-02-19
add report generation to inventory menu

yzhbankov:
Hello. I added report generation to inventory menu.
foods.client.view.html --> add dt-options='foodCtrl.dtOptions'
food.client.controller.js --> add dtOptions object which is determine datatable options
I hope this is enough for the issue. If not then i will work on it further.

closes#7

### DIFF
--- a/public/modules/food/controllers/food.client.controller.js
+++ b/public/modules/food/controllers/food.client.controller.js
@@ -10,6 +10,34 @@
 		// Copy food item for smart table
 		self.itemsCopy = [].concat(self.items);
 
+		// Add plugins into datatable
+		self.dtOptions =
+		{
+			dom: 'Tlrtip',
+			tableTools: {
+				sSwfPath: '/lib/datatables-tabletools/swf/copy_csv_xls.swf',
+				aButtons: [
+					"copy",
+					{
+						"sExtends":    "collection",
+						"sButtonText": "Save",
+						"aButtons":    [ {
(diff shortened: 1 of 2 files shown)


DealPete:
Hi yzhbankov,

You've implemented the right functionality, but I'm having some trouble getting it to work.
First of all, when I use my remote database (on mlab), it doesn't show any of the new buttons.
When I use my local database, I get different behaviour depending on the browser I'm using:

Safari: The Save button appears, but when I click on it there's nothing in the menu.
Chrome: The buttons don't appear when I select the Inventory menu, but if I reload the page they do appear, and work as expected.
Firefox: I can't get the buttons to appear.

In both Chrome and Safari, when the buttons do appear there isn't enough padding for the Items box and some of the text and buttons hug the side of the box.

yzhbankov:
I add empty <td> tags to table and the save buttons appears in page loading
The problem with Safari and Fire Fox in in the flash player. I refreshed flashplayer for firefox and safari and it all works fine. Without flash player the export buttons in other menus (Client Database, Volunteer Database, Donor, etc.) does not working too.

DealPete:
I didn't realize datatables made use of flash... alright, I've upgraded flash and it works now. The only remaining issue is the padding problem.

yzhbankov:
I changed table wrapper class and it looks better

DealPete:
perfect

PR #55, opened 2017-02-21
driver auth changes

Rusticman:
Hi Pete,

customer.server.controller.js ---> Changed ```hasAuthorisation``` function middleware so that it checks query variable ```driverDeliverPackage``` so it knows when driver is updating customer info.  This is where the code was breaking.

also changed the ```update``` function within the same controller to change this key -> ```driverDeliverPackage``` in the customer object back to ```false```.

customer.client.service.js  ---> changed this to send ```driverDeliverPackage``` as query string.

user-driver.client.controller.js ----> changed the factory service to ```CustomerUser``` as opposed to ```CustomerAdmin```, and passed the query string into the params here.

Let me know if there are any problems.

### DIFF
--- a/app/controllers/customer.server.controller.js
+++ b/app/controllers/customer.server.controller.js
@@ -203,7 +203,11 @@ exports.read = function(req, res) {
 exports.update = function(req, res) {
 	var customer = req.customer;
 
+	//changes field to false to stop unauthorized changes by driver
+  req.body.driverDeliverPackage = false;
+
 	customer = _.extend(customer, req.body);
+
 
 	// Adding fields not in the schema
 	var schemaFields = Object.getOwnPropertyNames(Customer.schema.paths);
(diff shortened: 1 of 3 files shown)


DealPete:
Hi Rusticman,

What's you've done works, but if I understand it correctly, it creates a security vulnerability. Any user who wanted to make changes to a client could set the `driverDeliverPackage` field on a client to `true` along with any other changes and it would be accepted by the server.

Rusticman:
There is another way:  The ```hasAuthorisation``` function, the server-side middleware checks if the volunteer has  'driver' as a role in order to allow changes to be made to a customer i.e change delivery date.  This means no query string in the server call.

Although I'm not sure if the original changes makes it more vulnerable, the original condition in the ```hasAuthorisation``` function works in tandem with the new one:

`if (req.customer._id !== +req.user.id && !driverDeliverPackage) {`

If the ```driverDeliverPackage``` variable isn't set (i.e ```undefined```), or if set to ```true```, the original condition still catches unauthorised requests.  Maybe I'm missing something  though, security isn't something I know a great deal about.  

Your call though, Peter.

DealPete:
In computer security you have to assume that your users are trying their best to sabotage things. With what you've described, one of your drivers could write some code that sets `driverDeliverPackage` to `true` on a customer, make some other changes, and submit those changes to the `customer/` endpoint. This is an unlikely scenario, but it does represent a security hole, because it gives a volunteer access to the database that only an admin should have.
I think what we need is a custom server route to be called from the user-driver controller. This route would confirm that the user is the driver assigned to deliver to the client in question, and only make those changes needed to mark a delivery as complete.

PR #56, opened 2017-02-22
fix bug - new Customer service

Rusticman:
customer.server.controller.js ---> compared the user object `id` property to the customer object `assignedTo` property to ensure the user is the driver of the client. 

customer.client.service.js ----> I created a new service called `CustomerDriver` as you requested.

user-driver.client.controller.js ---> changed the `CustomerUser` service to `CustomerDriver` service.

Let me know if there are any problems.

Cheers, Pete.


DealPete:
I see you've updated `hasAuthorization` to confirm that the user is assigned to deliver the to the client in question, but there's nothing I can see that prevents an authorized user from making arbitrary changes to the client. `CustomerDriver` still uses the `customer/:customerId` route, which calls `customer.update`, which calls `customer.save` to save to the database whatever changes the user has made to the client. We only want to change the one field `customer.lastDelivered`.

Rusticman:
Sorry Pete, I don't understand what you mean. 

>  but there's nothing I can see that prevents an authorized user from making arbitrary changes to the client. 

I've added a condition checking if the user changing the client's last delivery is his driver. I don't understand how a user can now make arbitrary changes that weren't possible before my changes.

> We only want to change the one field customer.lastDelivered.

I haven't changed the customer object in any other way.  

For clarification, could you please show me using pseudo code how you want this to work?

DealPete:
Well, how about I give you an example of what I'm concerned about.
Suppose the user Alice has been assigned as a driver to make a delivery to the client Bob. Alice is running the client-side code we send her when she goes to our website, and wants to play a prank on Bob. She changes
```
			}).forEach(function(customerOld) {
				var customer = new CustomerDriver(customerOld);

				// Update delivered date to this week
				customer.lastDelivered = beginWeek;
				// Add server call
				updatesInProgress++;

				customer.$update(function() {
```
to
```
			}).forEach(function(customerOld) {
				var customer = new CustomerDriver(customerOld);

				// Update delivered date to this week
				customer.firstName = "I love"
                                customer.lastName = "Donald Trump"

				// Ad server call
				updatesInProgress++;

				customer.$update(function() {
```
then clicks on "Mark Delivered". Is there anything on the server (which Alice has no control over) that prevents these changes from being made to Bob?

Rusticman:
Understood and nice example!

Rusticman:
hey Pete,
             I created a new service and route/controller.  An array populated with the relevant `client_id`s is sent to the controller server-side which handles the changes made to the `lastDelivered` property in the customer objects . This should stop any potential pranksters from changing the customer's details :)

DealPete:
Hi Rusticman,

You're definitely on the right track now - The `sendPackage` endpoint only has the power to update the `lastDelivered` field. However, there are still a couple of things to change:

1. Right now you have the client send the driver's user id to the server. There's no reason to do that, since the server keeps track of the user's id in its session middleware and exposes it as `req.user.id`.

2. `findPackage` should validate that the user in question is actually the driver for each of the clients in `arrayOfCustomers`.

Rusticman:
Hey Pete,
               I've made changes as requested.  I noticed Michael has said the code is changing to react.js.  Are you overseeing the migration?

PR #62, opened 2017-04-05
Reactify/user module

kenjiO:
I converted the forgot password, change password and change profile pages to react components.


kenjiO:
Closed because this was for another repository.

PR #63, opened 2017-05-26
Reactify

jspaine:


DealPete:
Looking good. I was able to get it up and running by following the "Quick Start Steps" in the readme. It is confusing, however, that when you type "npm run dev" it says "Application Started on Port 3000", but your really have to go to port 8080.
Also, the "admin/password" combination no longer works. Do you currently have a way to create an admin account?

jspaine:
Yeah, maybe we should log a message with the client url when compilation finishes.

It'll seed automatically if the database doesn't already exist or a parameter in the call to seedDb in server/config/express is set, if you still have an old version of it that's probably why. The schemas have changed so it's probably best to delete the old db or change the db name in the config.

DealPete:
Cool, I deleted the local db like you said and it works. The app is still a little buggy, but I can see that you must have put an immense amount of work in to get it to this point.

jspaine:
Nice, yeah there's quite a few bugs and half finished features!

PR #72, opened 2017-05-30
Use email addresses instead of usernames to identify accounts

kenjiO:


jspaine:
Nice, I think in the user authentication controller it should just say invalid user or password though. Let me know if you want to change it or I'll just merge this.

kenjiO:
Do you mean in the exports.signup method?

jspaine:
Ah it's signup, I'm not sure there's a better way to handle it at the moment then. Ideally we'd send an email with a confirmation token, then if it's an already signed up email send a password reset email instead I guess, to avoid revealing what accounts exist.

Not sure it's that urgent to fix though, could also use captcha, rate limiting...

PR #80, opened 2017-06-01
Fix bug when seeding with no addresses.csv file

kenjiO:
When this.file.length is 0 the range will still evaluate to [0] so I put a test to use [] when length is 0


jspaine:
Ah, sorry!

PR #82, opened 2017-06-02
Show server response and clear form on success

### LINKED ISSUE #73: Password reset page does not give feedback
The password reset page needs to give the user feedback for success and errors in the same fashion as on the signup page.  It should

- Indicate if the new passwords do not match
- disable the submit button if any fields are empty or if the new passwords do not match.
- If the server returns an error it should display the error
- If the server returns success it should display a success message and clear all the form fields

joshuahenson:
As I worked on this, I realized that the change password page and reset password page were similar so I worked on both. There are some minor differences in handling that I left between the two because I'm not sure which you prefer. I also widened the medium columns because the error/success messages were getting squished and I thought it looked odd. Widened ForgotPassword to match and fixed lint errors.

Let me know if you think anything needs to be changed.

Closes #73

### DIFF
--- a/client/modules/users/components/ChangePassword.js
+++ b/client/modules/users/components/ChangePassword.js
@@ -22,6 +22,12 @@ class ChangePassword extends React.Component {
     this.props.clearFlags()
   }
 
+  componentWillReceiveProps = nextProps => {
+    if (nextProps.success) {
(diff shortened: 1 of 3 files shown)


jspaine:
Looks good

PR #87, opened 2017-06-03
add websockets

jspaine:
It's kinda ugly, think it could be much cleaner if you allow another round trip to the server. Not sure it's worth trying to avoid it?


kenjiO:
I can't figure out what this does and what benefit is has.  Can you explain more what it is doing?

jspaine:
You can add a key to actions with a list of roles, when the api call gets to the server it intercepts the response and if it was successful it recreates the action and broadcasts it to connected users that match the role. So if you're signed in as admin looking at the customer list and a new customer applies you'll get the customer/SAVE_SUCCESS action and they'll show up in the table without you needing to refresh.

Probably not that useful in most cases but kinda nice to have and I thought it could be useful for tracking deliveries.

kenjiO:
I think it would be better to hold off on adding this until sometime in the future when the application is in a more release ready state.  I'm afraid it will add more complexity and make it harder to fix the current issues we have.  What do you think?

jspaine:
Yeah maybe. The code around it isn't going to change much so it should be easy to merge in any time. It's also easy to disable by commenting out the on connection users.push line on the server though.

kenjiO:
OK, we can put it in now.  When I ran the tests it broke most of them.  Can you fix that?  Also in client/app.js line 32 I think we should have it not be hard coded to localhost:3000

jspaine:
:o Managed to clean it up quite a lot and got it working for most of the things I think it would be useful for. Still not that happy with the approach though, maybe leave it here a bit longer.

jspaine:
This is ok I think, there's nothing on the client except to open the socket, keep a ref in the api middleware and send the id in a header. What to send is defined in the server routes.

PR #110, opened 2017-06-13
more confirmations

### LINKED ISSUE #94: Page should re-direct after deleting a customer
When logged on as an admin if you view a customer and click the Delete button it deletes the customer but it still stays on the same page.  It should go back to the customer list or show some indication that the customer has been deleted.

### LINKED ISSUE #98: Volunteer application form modal not appearing correctly
When you initially save the form or if you save edits to the form the modal indicating do you want to leave with unsaved changes comes up.  This modal should come up when you cancel the form instead.

### LINKED ISSUE #107: Add delete confirmation dialogs
When clicking the delete button for customers, donors or food items it should prompt to confirm before deleting in case the use accidentally clicks the delete button.  If possible use the same style and code as the confirmation for the warning about leaving unsaved data when editing pages.

jspaine:
I didn't add to food item delete, not sure it's needed there?

closes #107, closes #98

### DIFF
--- a/client/app.js
+++ b/client/app.js
@@ -46,8 +46,8 @@ function render(Component) {
 
 if (module.hot) {
   module.hot.accept('./Application', () => {
-    // const Next = require('./Application').default
-    render(Application)
+    const Next = require('./Application').default
+    render(Next)
   })
 }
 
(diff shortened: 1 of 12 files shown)


client/app.js
```diff
@@ -46,7 +46,8 @@ function render(Component) {
 
 if (module.hot) {
   module.hot.accept('./Application', () => {
-    render(Application)
```
jspaine, line 49: This was supposed to work in webpack 2, but it didn't seem to for me so I changed it back to the old way. Did you notice anything?

kenjiO: I noticed hot reloading wasn't working at times.  This fixed it for me

PR #135, opened 2017-06-16
allow passing secrets as env vars

jspaine:


jspaine:
Do the sendpulse credentials work for you Kenji? I get something about failed to authenticate by 2 methods.

PR #139, opened 2017-06-20
Additional server tests for signup and signin

kenjiO:


server/lib/questionnaire-validator.js
```diff
@@ -4,9 +4,10 @@ import Questionnaire from '../models/questionnaire'
 import validate from '../../common/validators'
 
 export default function getValidator(identifier) {
-  if (process.env.NODE_ENV === 'test') return () => true
 
   return async function validator(fields) {
+    if (fields.length === 0) return true
```
jspaine: If there's a required field in the questionnaire and you submit no fields this will pass when it should fail won't it?

kenjiO:
Here is a revised one.  I added questionnaires to the test database to get the validators to work in the test environment.

jspaine:
Why though? I don't see the benefit to having to create empty questionnaires in the client tests. The validator can still be unit tested by rewiring `process.env.NODE_ENV` in the test, or, probably better, adding a shouldValidate parameter to the getValidator function.

kenjiO:
I would rather keep `if (process.env.NODE_ENV === 'test') return () => true`  out of the main code and make the tests more complex.  It will also make the tests more realistic too.

jspaine:
Ok but its just sort of testing the validators dependency on questionnaires (by giving it so it doesn't crash), and that could be easily and explicitly tested in a unit test.

PR #145, opened 2017-06-30
Keep a list of packed packages

kenjiO:
This is for the issue of being able to undo marking a customer as packed.  I think it is best to keep track of the packed packages to do this.  So this is the part that just handles keeping track of what is packed.  Next week I'll work on the feature of undoing a packed package.  Some time in the future it might be good to allow users to add other items to packages and specify the quantity of items.


jspaine:
Looks good

Issue #144, opened 2017-06-26
HTML <title> tag is missing


kenjiO:
There is no <title> tag in the head section of the html page.   The html page template is in `client/index.html`.  It should have a title tag and the title should be dynamically set to what the use enters in the general settings page organization field.

jspaine:
closed in #162

PR #162, opened 2017-07-06
Fix title missing

kevinnorris:
In reference to issue #144 

As far as I am aware in order to include a ```<title>``` tag in the html page that updates, one would have to add server side rendering to the application. 

I added a quick fix. The top most component that is connected to the store ```client/Application.js``` does a check on update and changes the title if it differs from the```settings.organization``` value in the store.

### DIFF
--- a/client/Application.js
+++ b/client/Application.js
@@ -12,8 +12,7 @@ import ServerError from './modules/core/components/errors/ServerError'
 const mapStateToProps = state => ({
   user: selectors.user.getUser(state),
   loading: selectors.user.fetching(state),
-  error: selectors.user.error(state),
-  settings: selectors.settings.getSettings(state)
+  error: selectors.user.error(state)
 })
 
 const mapDispatchToProps = dispatch => ({
@@ -46,10 +45,6 @@ class Application extends Component {
     window.dispatchEvent(new Event('resize'))
     if (user && user.roles.find(r => r === ADMIN_ROLE)) {
       $('body').removeClass('layout-top-nav')
-    }
-
-    if (document.title !== this.props.settings.organization) {
-      document.title = this.props.settings.organization
     }
   }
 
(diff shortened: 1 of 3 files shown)


jspaine:
I think the most popular is [react-helmet](https://github.com/nfl/react-helmet). Ideally there'd also be an `app` reducer in `client/modules/core` that could set the title and listen to router location changes. Extra title parts could be got from the title attribute of the active menu item and maybe route params like :donorId.

Let me know if you want to try any of that, want more details or just want me to merge this.

kevinnorris:
Could you extrapolate on the extra title parts you are thinking of? 
Having the title change a lot while you are on the same page might be a bit off-putting for users. 

I think having a title based on the page you are on is a good idea.
How would you envision that to look? Maybe something like ```pageName | organization```?

Do you think it would be better to implement it as an ```app``` reducer as you previously mentioned and just use ```document.title```, or throw in ```react-helmet``` and maybe have a helmet for each page including a title and a description . Is it worth putting in another dependency and enforcing every page to include a helmet component?

jspaine:
I was thinking like (eg. for editing/viewing a customer): {customerId/name} - Customers - Foodbank App

or just: Customers - Foodbank App

I think there should be a component (it can just be the Application component or make another one maybe in modules/core) that sets the title from redux state, it can use document.title, helmet or whatever. Then at worst you just need to dispatch a setTitle action from components.

But ideally it would be automatic so you don't need to change all the routable components. I was thinking you could pull it from the menu items (`modules/core/reducers/menu`) but there are pages that aren't linked directly from the menu. Maybe it could work like:
- Try to find a menu item that matches the url exactly, then use the menu items title attribute
- Otherwise split the url by '/' and use the capitalized first part as the title
- (Optionally) If there's a customerId, donorId etc. router param add that too or, better, use it to look up the name

jspaine:
Nice, it looks good! Couple of little things:

There's a lodash capitalize function that capitalizes the first letter of a string,
There are getOne selectors that take `state => id => customerWithId`

so you can do in mapStateToProps: `getCustomer: state => selectors.customer.getOne(state)`
then call the function in the component `const customer = this.props.getCustomer(id)`

kevinnorris:
That does make it a lot cleaner 👍

jspaine:
Nice :smiley:

PR #165, opened 2017-07-06
Reply with 400 when client submits bad json

kenjiO:
If a user sends a bad json request such as  **POST /api/auth/signin  {"email": "password":}** the server responds with 500 'Something went wrong on the server'.  This changes it to reply 400 and informs there is a problem with the json the user submitted.

### DIFF
--- a/server/lib/errors.js
+++ b/server/lib/errors.js
@@ -78,12 +78,13 @@ export default function getErrorMessage(err) {
   }
 
   //This catches when body-parser encounters bad JSON user data in a request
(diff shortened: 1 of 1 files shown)


jspaine:
What's in err.message and err.body though? The error pretty much means there's a bug in the client, I don't think we need to say much more than 'oops something went wrong, try refreshing' or something.

kenjiO:
How about like this.  For development it is helpful to get a more detailed message.  
err.message comes out to be something like "Unexpected token x in JSON at position 1"
err.body is the JSON string from the request body

jspaine:
Yeah that sounds fine.

Issue #106, opened 2017-06-12
Category changes when editing a food inventory item


kenjiO:
When I go to the food inventory page, click the edit button for a food item, then edit the food name the category will change to the first category and the category field will be outlined in red.  Editing the food name should not change the category

jspaine:
closed in #170

PR #170, opened 2017-07-07
Fix inventory category updating incorrectly

kevinnorris:
#106 

Name changes were reseting the categoryId to ```""```.  Changed to set it to the previously set categoryId.


jspaine:
Good job, I added a few more inventory issues in #171 if you want to have a go.

Issue #171, opened 2017-07-07
Food Inventory


jspaine:
Few small issues with inventory page:
- New item modal 
Should show errors only on touched fields
Should autofocus name field
On select food from autosuggest should focus quantity
(low priority) up/down arrows in autosuggest list skip an item

- Categories
Edit should autofocus category name input

-General
Enter should submit, escape cancel

ronaldblanco:
Hello Jspaine;

What else do you need with this issue?, is it close?
If I can help you here let me know.

jspaine:
Hi, everything that's not checked. You're welcome to look at it. The category stuff should be pretty easy to start with.

sherwinmina:
Hi Jspaine,

I worked fixed two of the issues on Categories. I will also work on the New Item Modal.

Thirdoptics:
Fixed the last issue in Categories with the autofocus on edit with [Pull Request #268](https://github.com/freeCodeCamp/pantry-for-good/pull/268).

Thirdoptics:
Addressed 

> On select food from autosuggest should focus quantity

 in [Pull request 269](https://github.com/freeCodeCamp/pantry-for-good/pull/269)

Thirdoptics:
Addressed 

> Should show errors only on touched fields

in [Pull request 270](https://github.com/freeCodeCamp/pantry-for-good/pull/270)

PR #177, opened 2017-07-08
Update Inventory Modal

kevinnorris:
Partial solution to #171 

* Added form submitting on enter
* Form exit with ```esc``` seems to work
* New item, name field autofocus
* Edit item, category field autofocus
* Error warnings only show after change

Since you have the add/update button disabled until all fields are validated, I think it might be unintuitive if the user has filled in some of the fields, can see no validation errors, and yet the button is still disabled. Therefor I set the form to show validation once the user starts to interact with it.

An alternative approach could be to show no validation errors and show an active button. Then only after the user tries to submit does the form show validation errors if any occur.

As for autosuggest, is there an easy way to differentiate between autosuggest and user input?

### DIFF
--- a/client/modules/food/components/inventory/FoodItems.js
+++ b/client/modules/food/components/inventory/FoodItems.js
@@ -17,7 +17,8 @@ class FoodItems extends React.Component {
       editModalFood: undefined,
       modalInputFields: { name: "", categoryId: "", quantity: "" },
       validInput: false,
-      searchText: ""
+      searchText: "",
+      hasBeenChanged: false
     }
   }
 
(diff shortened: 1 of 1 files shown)


jspaine:
The way the other forms work is all fields start off untouched, if you change one and blur it it gets marked touched. If you submit the form all fields get marked touched and any invalid prevent the actual submit. Touched && invalid fields show errors. 

You can probably skip the blur for this and just mark touched on change.

Looks like there's an [autosuggest onSelect prop](https://github.com/affinipay/react-bootstrap-autosuggest/blob/gh-pages/apidocs/Autosuggest.md#onselect)

Issue #97, opened 2017-06-06
Do we need the food schedule items start date?


kenjiO:
I don't know how useful this data is.  If we do keep it, it should be formatted in a more readable way and fix the functionality to edit it.

jspaine:
Yeah it's pretty clunky. Also frequency seems to actually mean interval.

How about an input to select a single start date and change that column to be offset in weeks. If you wanted to send eg. two items alternate weeks you'd set their interval to 2 and first offset to 0, second offset to 1. If the interval is weekly offset should be disabled, if interval is 2 (fortnightly) offset should be 0 or 1... So max(offset) should be interval - 1.

Maybe it would be less confusing with words too, eg offset = second and fourth week, interval = fortnightly or offset = third week, interval = monthly. Wouldn't work for 3 week intervals though.

yaseenshaik:
To solve some confusion, I think we need to  
* show start date as a single date 
* change frequency to interval in weeks or just show "Every" label. 
* get actual days of the week 

For example:
* Person 1 wants to donate every Friday: `days: [Friday], interval: 1` ("Every week")
* Person 2 wants to donate Monday and Wednesday every fortnight: `days:[Monday, Wednesday] interval: 2` ("Every Fortnight")

Similarly `interval: 4` would be "Every Month". This will cover all cases.

jspaine:
It's a schedule for packing items for delivery, it's assumed there's one delivery a week. It could be nice if it was more flexible but it's another issue I think.

I'm not sure it needs any dates if you think of the interval as just every n deliveries., The start date column could be replaced with the offset into the cycle, and you could have a single control at the top to skip forward or backward in the cycle. I think this would make it easier to work with customizable delivery dates too.

yaseenshaik:
Ok now I'm confused. Why do you assume only one delivery per week? Isn't
the list showing food items and not customers?

If it's just a schedule for packing, what does frequency stand for? And
what's the need for having a packing schedule. Isn't there something better
already?

jspaine:
Yeah, sorry!

The app always assumed one delivery a week and we haven't changed that. I guess it's the most common scenario too.

The idea is that some items you might not want to send so regularly, so maybe non-perishable stuff like rice and pasta you could send every other week, while perishable stuff like milk you'd send every week.

PR #179, opened 2017-07-08
Fix/frequency validator

### LINKED ISSUE #176: Schedule items frequency should not allow negative numbers
On the Food Schedule page the Frequency column values can be set to negative numbers.  The application should not allow values less then 0.

sherlyc:
added a validator for frequency column at food delivery, do not allow negative and decimal numbers.
closes #176

### DIFF
--- a/client/modules/food/components/Schedule.js
+++ b/client/modules/food/components/Schedule.js
@@ -46,6 +46,13 @@ class Schedule extends Component {
     return category && category.category
   }
 
+  frequencyValidator = (value) => {
+      if(value < 0) {
+          return 0;
+      }
+      return true;
+  }
+
   render() {
     const {loading, saving, error} = this.props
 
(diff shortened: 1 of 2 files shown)


jspaine:
Thanks, it would be nice if you could just set min and step props on the number field but react-bootstrap-table makes it a bit difficult. We're also not really sure the best way to do this page. There are some ideas in #97 if you want to do more on this.

sherlyc:
is it possible to add min and step props on bootstrap-table? I have looked through its documentation, didn't find anything useful.

jspaine:
The second last example [here](https://allenfang.github.io/react-bootstrap-table/example.html#celledit) with customEditor would do it, or I made a PR that lets you do `editable={{type: 'number', attrs: {min: 0}}}` but I'm not sure it'll be merged.

PR #190, opened 2017-07-31
Fix editing Food Schedule Frequency and Start Date

### LINKED ISSUE #96: Can't edit Food Schedule frequency column
When logged in as an admin if you go to the schedule page there is no way to edit the food frequency column.  However if you edit the start date and then click on the frequency it will increment the start date.  Clicking on the frequency should edit the frequency and not change the start date.

yaseenshaik:
Editing start date and frequency was not possible in firefox (and other non-chrome browsers?). This PR aims to fix both. Since this is also my first PR, and I would like to do more, I would like your feedback and a project overview. (I looked at the developer's guide and its outdated. Even a general idea of how everything is organized and a list of supporting libraries would be great.) 

closes #96

### DIFF
--- a/client/modules/food/components/Schedule.js
+++ b/client/modules/food/components/Schedule.js
@@ -1,8 +1,10 @@
 import React, {Component} from 'react'
 import {connect} from 'react-redux'
-import {utc} from 'moment'
+import moment, {utc} from 'moment'
 import {BootstrapTable, TableHeaderColumn} from 'react-bootstrap-table'
 import 'react-bootstrap-table/dist/react-bootstrap-table.min.css'
+import DayPicker from 'react-day-picker'
+import 'react-day-picker/lib/style.css'
 
 import selectors from '../../../store/selectors'
 import {loadFoods} from '../reducers/category'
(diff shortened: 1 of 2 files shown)


jspaine:
Hi, I updated the developers guide, hope it's ok for a rough outline. Let me know if there's anything else.

That's strange it's not possible to edit as a number, I think it's a react-bootstrap-table bug. It works if you disable blurToSave. 

Editing the date works on firefox for me, but it's the same problem as on chrome the week increments even if you don't change it. I'd rather not have to use a date-picker to fix it (there's nowhere else in the app that needs one) but I think this page needs a redesign anyway (see #97) so I'll merge this as a temporary solution.

yaseenshaik:
Thanks for accepting my PR and updating the dev guide.

HTML5 inputs are not supported equally across all browsers. Before this patch I couldn't get the frequency column to work on FF 54 on Linux. (Maybe there is a bug in react-bootstrap-table that needs to be fixed?)

I never wanted to go with a special date picker component as well, but using one not only works across all browsers, but also a consistent experience.

I agree that food schedule needs rework in order to make it less confusing.

PR #193, opened 2017-08-03
Update server packing controller pack method to use the mongoose model validator

### LINKED ISSUE #192: Invalid data submitted to create new packages causes an inconsistant state
When multiple packages are submitted to POST /api/packing, it will save the packages that are valid but if it encounters a package with invalid data it throws an error without going on to update the customer last packed dates and food item quantities of the packages it has already saved.  Also it doesn't save any remaining packages with valid data.

kenjiO:
and do all validation before saving

fixes #192.  Any invalid data will cause the whole operation to fail without saving anything.

### DIFF
--- a/server/controllers/packing.js
+++ b/server/controllers/packing.js
@@ -25,41 +25,54 @@ export default {
     if (!Array.isArray(packages)) {
       throw new BadRequestError('Request body must be an array')
     }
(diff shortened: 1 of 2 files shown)


jspaine:
Looks good.

Issue #199, opened 2017-08-04
Set the copyright year to current year


kenjiO:
In the footer to the website, the **copyright 2016** is hard coded.  Change it so the year is set programmatically to the current year.  The footer is a react component in the file
`client/modules/core/components/Footer.js`

dmcfaddengalway:
I'd like to this issue up! Let me know if this is okay.

dmcfaddengalway:
Made the required changes and have made a PR, followed the Contributing guidelines and such

dmcfaddengalway:
Just made the changes, let me know if this is more of what you want.

kenjiO:
PR #201 closed this issue

PR #201, opened 2017-08-04
Fixed copyright year

dmcfaddengalway:
Changed the copyright year on client/modules/core/components/Footer.js fixes

### DIFF
--- a/client/modules/core/components/Footer.js
+++ b/client/modules/core/components/Footer.js
@@ -8,7 +8,7 @@ const mapStateToProps = state => ({
 
 const Footer = ({settings}) =>
   <footer className="main-footer">
-    <strong>Copyright &copy; 2016&nbsp;
+    <strong>Copyright &copy; 2017&nbsp;
       <Link to="/">{settings && settings.organization}</Link>.
     </strong>
     &ensp;All rights reserved.
--- a/client/modules/core/components/Footer.js
+++ b/client/modules/core/components/Footer.js
@@ -6,8 +6,8 @@ const mapStateToProps = state => ({
   settings: state.settings.data,
 })
 
-var date = new Date();
-var year = date.getFullYear();
+var date = new Date()
+var year = date.getFullYear()
 
 const Footer = ({settings}) =>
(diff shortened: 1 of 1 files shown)


jspaine:
Hi, thanks but the issue was to make it automatic. So you'd need to find the date, get the year and show that.

dmcfaddengalway:
Oh I can definitely do that for you! Mind if I continue with this issue

jspaine:
Looks fine, but try and stick to the same style as the rest of the code - use const if possible or let, and no semicolons at the end of lines if possible.

dmcfaddengalway:
Will fix right now!

Thank you and have a nice day,
     Daniel McFadden
     +353 (083) 8844332 (IRL)
     +1 (215) 350 4830 (US)
     dmcfaddengalway@gmail.com (primary)
     dmcfaddenhgp@gmail.com (secondary)

jspaine:
Sorry for the delay, been moving.

And sorry but I meant use `const` if possible or `let` instead of `var` for variable declarations. It doesn't make any difference in this case but it's more consistent and generally a bit safer and more informative (this thing is constant, this is mutable). And the remaining `var`s are kind of a marker for old code that should maybe be updated.

Also there's no need for double curly braces around `year` in jsx.

dmcfaddengalway:
@jspaine No worries! I did use `let` as I was under the impression that `const` would make its value not changable. But I will look into this more as I must have been confused. Will also change from `let` to `const` for you and resubmit.

I will fix the double curly braces around year. My bad, new to React.

jspaine:
Awesome, thanks!

I didn't see let, maybe I missed it. But the value isn't going to change so const is right. It'll be set when the file is imported and stay the same until the app is closed or refreshed. So it won't update until a refresh if the app is kept open over the new year but I don't think that matters.

closes #199

dmcfaddengalway:
@jspaine Looking back at it, it appears I must have actually put `var`, not `let` -- muscle memory. 

Oh okay, understand now. Thanks!

PR #203, opened 2017-08-11
Add Units of distance and Corrency on settings

### LINKED ISSUE #159: Units (km/mi, $ etc)
Admin should be able to set currency and distance units.

ronaldblanco:
Add Units of distance and Currency on settings model and RouteSummary show distances base on the unit set (only between km -kilometers- and mi -miles-)
The application and all changes where locally tested.
 closes #159


jspaine:
I don't know what you're doing wrong but this is a mixture of the old version of the app and the new one. There are 2000 changed files!

If you can fix that the actual code you've written looks ok and I'll give you some suggestions.

I'd say keep the project folder on your computer, delete your fork on github and make a new fork. Then clone it into a different folder on your computer, **check you're on the staging branch**, and copy the changes you made into there.

Issue #123, opened 2017-06-15
Add propTypes to components


jspaine:

ml9951:
I'm interested in tackling this one.  Is there any opposition to using Flow instead of PropTypes?  I think `flow` would allow for more expressive type checking and allows us to do it statically [[1](https://flow.org/en/docs/react/components/)].

jspaine:
I'm ok with that, would be useful lots of other places too.

Don't think you have to do every component btw, if you can get it set up and do a good handful that would be great.

PR #217, opened 2017-08-18
Adding flow and flow-typed definitions

ml9951:
This PR adds the necessary infrastructure to get `flow` running.  I plan on implementing `prop` types ( #123 ) in the near future, but thought that other people could make use of it elsewhere.  In particular, I think that typing `redux` actions is an excellent use case for `flow`.  It might be good to add `flow` to the `pre-commit` hook that runs `eslint` and the testing suite.


jspaine:
Nice, but do you have to include all the types for node modules? Can they be generated in a postinstall step or something?

> It might be good to add flow to the pre-commit hook that runs eslint and the testing suite.

Yeah sounds good.

ml9951:
I agree it's a bit clunky to check them all in, that's just what's recommended in their README [[1](https://github.com/flowtype/flow-typed)]

> All you have to do when you add one or more new dependencies to your project is run flow-typed install. This will search the libdef repo and download all the libdefs that are relevant for your project and install them for you. After that, simply check them in and be on your way!

I like the idea of adding them in a postinstall step.  I'll give that a shot and get back to you.

ml9951:
I'm having an issue with the type defs for Enzyme [[1](https://github.com/flowtype/flow-typed/issues/1158)].  I submitted a pull request to fix them [[2](https://github.com/flowtype/flow-typed/pull/1159)] but it has yet to be merged.  If we were to just checkin the typedefs, I could manually fix them.  What do you think?

ml9951:
No problem.  I recently realized that I need to make a minor change to properly deal with importing CSS modules, so that should get changed before this is merged anyways.

jspaine:
CSS files? What's the problem, just curious.

ml9951:
It's a very minor problem with a pretty trivial fix [[1](https://github.com/ArnholdInstitute/ATLAS/blob/master/package.json#L87)].  You basically just want to either ignore them, or map them to an empty module.  Otherwise, `flow` will try and typecheck them and inevitably fail if they are imported in a javascript file.  The same goes for importing images and other non-javascript files.

jspaine:
Ah, I had to use [ignore-styles](https://github.com/bkonkle/ignore-styles) for tests, not sure you can use that?

ml9951:
Yeah, I'm not sure if that would work for `flow`.  In any case, it's about 3 lines to accomplish the same thing.  This should be good to go on my end if you decide you'd like to commit the typedefs.

kenjiO:
I've never used flow before so I'm not familiar with it.  I just read up on it a little bit today.  I'm fine with using it with the flow-typed libraries and built in checking.  I do have some concern if we want to start adding types to the code that it will make it more difficult for those who are not familiar with flow to contribute to the project.

ml9951:
> Is there an easy way to ignore errors from a file, but still pass it through flow so all the types don't throw syntax errors? Then someone could just disable it and it could be fixed later.

There is.  You can just remove the `//@flow` comment from the top of a file and `flow` will ignore it.

> when many things are typed it could be confusing

Certainly there is a small learning curve to getting started with flow, but I also believe that the addition of types can dramatically improve the readability of a codebase and provides implicit documentation that is useful for newcomers.

jspaine:
Sounds good.

kenjiO:
OK, I'll go ahead and merge it

Issue #132, opened 2017-06-15
Driver and food bank locations should be updated when the address is changed


jspaine:
This is implemented for customers as a mongoose pre-save hook and can probably be done similarly. Note the `submitValidate` meta tag is set in `saveCustomer` in the customer reducer to enable redux-form submit validation.

ronaldblanco:
Hey I think I can help you here.
Give me a little more of detail.

jspaine:
Hi, look at the customer model and you'll see the pre-save hook with the geolocation code. There needs to be something similar in the settings and volunteer (for drivers) models.

ronaldblanco:
I will be working on this.

ronaldblanco:
Hey I was checking the "Route Assignment" option on the staging branch and the bottom "Suggest Route" it is not working!
So, the objective of this issue is to have the location information (lat and lng) also on Volunteer and Settings models right??

jspaine:
I can't check it at the moment, but don't see any reason it would be broken. There's a limit of 23 stops (21 customers) for calculating a route with google though, maybe that's the problem?

They have location info because the database is seeded with it, it's just not updated when addresses are changed and it needs to be.

ronaldblanco:
Hey I did a pull request to this issue; check it and let me know if something must be change.

Thanks.

ronaldblanco:
Hello Jspaine;
Is it a problem with the solution to this issue?

jspaine:
Hi, I didn't actually revert it, I thought about it because the linter was failing due to unused variables and a forgotten import of ValidationError in `pre-save-functions.js`, so nobody could commit because it requires that it passes. And now I'm not sure how to remove that commit, or even where it is - it doesn't seem to exist on the staging branch.

I fixed the errors and tidied it up a bit instead, you can see what I did here - https://github.com/freeCodeCamp/pantry-for-good/pull/228

I know the tests weren't working on windows, but it should be fixed now. So if you disabled the precommit hook can you reenable it? It's also helpful to have eslint run in your editor so you can see the errors live, see if there's a plugin for yours.

ronaldblanco:
Ok, I will be searching for a eslint module or pluging.

ronaldblanco:
I am sorry for the inconvenience.

jspaine:
It's fine, wasn't hard to fix :smile:

PR #224, opened 2017-08-28
(#132) Driver and foot bank location should be update when the addres…

ronaldblanco:
…s is changed.
Created the pre save function on the volunteer model to update the location and created the file "settings-location-update" with a function for the update of the location base on the address change on settings model.
All changes were tested on a local running app to ensure good stability and work.
closses #132

### DIFF
--- a/server/controllers/settings.js
+++ b/server/controllers/settings.js
@@ -5,6 +5,30 @@ import {ADMIN_ROLE, volunteerRoles} from '../../common/constants'
 import config from '../config'
 import Settings from '../models/settings'
 
+import nodeGeocoder from 'node-geocoder'
+import {fieldTypes} from '../../common/constants'
+import {getFieldsByType} from '../lib/questionnaire-helpers'
+import locationSchema from '../models/location-schema'
+import {ValidationError} from '../lib/errors'
+
+// Initialize geocoder options for pre save method
+const geocoder = nodeGeocoder({
+  provider: 'google',
+  formatter: null
+})
+
+/**
+ * Hook a method to construct the geolocation of the address
+ */
+async function updateLocationSettings(address, id, next){
+  const [result] = await geocoder.geocode(address)
+  if (result && result != undefined && result != null && result != {}) {
+    const {latitude, longitude} = result
+    await Settings.findByIdAndUpdate(id, {location: {lat: latitude, lng: longitude}})
+    return 0
+  } else if (!result) return next(new ValidationError({address: 'Address not found'}))
+}
+
 
 export default {
   async read (req, res) {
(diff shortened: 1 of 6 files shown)


server/lib/settings-location-update.js
```diff
@@ -0,0 +1,33 @@
...
+
+  const query = Settings.findById(id)
+  const settings = await query.select('+gmapsApiKey +gmapsClientId')
+
+  const address = settings.address
+
+  const [result] = await geocoder.geocode(address)
+
+  if (!result) return next(new ValidationError({
+    fields: Object.assign({}, ...addressFields.map((field, i) => ({
```
jspaine: The settings model doesn't have a 'fields' key, what key should the error be on in this case?

server/lib/settings-location-update.js
```diff
@@ -0,0 +1,33 @@
...
+      [field.meta]: i === 0 ? 'Address not found' : ' '}
+    )))
+  }))
+
+  const {latitude, longitude} = result
+  await Settings.findByIdAndUpdate(id, {'location': {'lat': latitude, 'lng': longitude}})
```
jspaine: Don't quote object keys unless you need to

server/models/volunteer.js
```diff
@@ -80,6 +84,36 @@ var VolunteerSchema = new Schema({
...
+  // Initialize geocoder options for pre save method
+  const geocoder = nodeGeocoder({
+    provider: 'google',
+    formatter: null
+  })
+
+  /**
+   * Hook a pre save method to construct the geolocation of the address
+   */
+  VolunteerSchema.pre('save', async function(next) {
```
jspaine: This could be extracted into a separate file because it's used in both volunteer and customer models

server/controllers/settings.js
```diff
@@ -5,6 +5,8 @@ import {ADMIN_ROLE, volunteerRoles} from '../../common/constants'
 import config from '../config'
 import Settings from '../models/settings'
 
+import {updateLocationSettings} from '../lib/settings-location-update'
```
jspaine: Why can't this be a pre-save method too? Otherwise I'd just make it a function in this file to take the address and return the location, then add the location to the request body before saving. Then you don't need to lookup and save twice.

server/lib/settings-location-update.js
```diff
@@ -14,20 +14,13 @@ const geocoder = nodeGeocoder({
...
-
-  if (!result) return next(new ValidationError({
-    fields: Object.assign({}, ...addressFields.map((field, i) => ({
-      [field.meta]: i === 0 ? 'Address not found' : ' '}
-    )))
-  }))
-
-  const {latitude, longitude} = result
-  await Settings.findByIdAndUpdate(id, {'location': {'lat': latitude, 'lng': longitude}})
+  if (result) {
```
jspaine: I meant there should still be a ValidationError if no result is returned, but there are no `fields` in the settings model to show it on. So the error should point to something that is in the settings model. Where would you want to show the error on the form if the address doesn't geocode?

ronaldblanco:
Hey take a look now, I did follow your suggestions.

ronaldblanco:
I also did the changes in the reducers as you toll me.

server/controllers/settings.js
```diff
@@ -5,6 +5,28 @@ import {ADMIN_ROLE, volunteerRoles} from '../../common/constants'
...
+  provider: 'google',
+  formatter: null
+})
+
+/**
+ * Hook a method to construct the geolocation of the address
+ */
+async function updateLocationSettings(address, id){
+  const [result] = await geocoder.geocode(address)
+  if (result && result != undefined && result != null && result != {}) {
```
jspaine: You need something like this to show the error on the form:
```js
if (!result) return next(new ValidationError({x: y}))
```
where x is the field of the model that has the error and y is the error message. Then when there's no geocode result the form will show field x in red with the error message y.

jspaine: You need something like this to show the error on the form:
```js
if (!result) return next(new ValidationError({x: y}))
```
where x is the field of the model that has the error and y is the error message. Then when there's no geocode result the form will show field x in red with the error message y.

ronaldblanco: Is it this OK??
if (!result) return new ValidationError({address: 'Address not found'})

I do not know how to used this on the controller because I am returning a res.json(settings).

jspaine: It has to be like I wrote, `return next(new ValidationError(...`

The router calls controllers with `req, res, next` params, so you can just add the `next` parameter so you can use it.

It should be, if there's no result you return from the controller without saving the model, and call `next` (the next express middleware) with an `Error`. Then the error handler middleware will catch it and send it to the client.

If there is a result you save the model and respond with the updated model.

ronaldblanco:
Excellent, thanks for your help.
Check now; it was working perfect!

jspaine:
Thanks!

PR #226, opened 2017-08-31
Add automatic session expiration (#22)

### LINKED ISSUE #22: Implement session timeout
Implement automatic session expiration, in accordance with OWASP. See https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Session_Expiration

yaseenshaik:
* Automatically expire session if idle for an hour. see `server/config/env/all.js:14`.
* Also fixed a bug that allowed user to set themselves as active even if they're rejected. (I was the one who introduced the bug with my last PR.)

Closes #22

### DIFF
--- a/client/modules/customer/components/CustomerEdit.js
+++ b/client/modules/customer/components/CustomerEdit.js
@@ -97,7 +97,7 @@ class CustomerEdit extends Component {
                 initialValues={toForm(customer, questionnaire)}
               />
             }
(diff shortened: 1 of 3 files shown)


jspaine:
Thanks!

> Also fixed a bug that allowed user to set themselves as active even if they're rejected.

The server should block this too, i'll add an issue.

PR #231, opened 2017-08-31
Add page to list all users

kenjiO:
James, can you look at this and figure out what is wrong with the new "User Accounts" page I created.  When I go to that page the UserList component keeps mounting and unmounting in an endless cycle.  I can't figure out what is causing it to unmount.  Thanks!


jspaine:
Yeah that's weird, if you change it to `if (!this.props.loading && !this.props.users)` it works.

I'll try and figure it out tomorrow.

jspaine:
I think you should make a separate reducer and keep that one for the current user. A lot of the app listens to it and setting and clearing fetching looks like a user's just logged in. The crud reducer helpers make it pretty easy anyway.

Maybe there should even be a separate `client/modules` for this. (If you don't have something that does it, I've got a patched refactoring thing for updating imports after renaming, so just call it whatever for now and I can change it later)

kenjiO:
I created a separate reducer and the problem went away.  I named the new reducer 'user' and renamed the old 'user' reducer to 'auth'.  Do you think it would be good to move the set_profile functionality from the 'auth' reducer to the 'user' reducer?

jspaine:
Cool. I'd leave it there and filter out the current user from the list of users, otherwise you'd have to keep the user in both reducers wouldn't you?

kenjiO:
I finished it.  Can you merge it if it looks OK to you.  I would keep it as 3 separate commits since there are lots of changes

Issue #232, opened 2017-09-06
/users/profile page getting 403 errors for users without a role


kenjiO:
All user accounts used to be created with an initial role but now user accounts are not assigned a role initially.  For any user without a role trying to go to the /users/profile page will get a 403 unauthorized error.

In client/modules/user/UserRouter.js the route for 'profile' should test for a logged on user and not for a valid role.

davycheung:
Added a PR #237

nodox:
Probably should close this issue.

PR #237, opened 2017-09-09
user login or redirect for change password and edit profile pages, no…

davycheung:
… need to check roles for those

Addresses issue #232


jspaine:
Thanks!

PR #238, opened 2017-09-11
fixed Category.js issues on enter submit, autoFocus and escape/cancel

sherwinmina:


client/modules/food/components/inventory/Category.js
```diff
@@ -14,6 +14,17 @@ class Category extends React.Component {
...
   }
 
+  onEnterSubmitEdit = e => {
+    if (e.key === "Enter") {
+      this.setState({ showEdit: false })
+      this.props.onItemEdit(this.props.id, this.state.editedName)
+    }
+
+    if (e.key === "Escape") {
+      this.setState({ showEdit: false })
```
jspaine, line 24: Should you reset `state.editedName` here or in `onClickShowEdit`? If you change the name, cancel, and edit again does it show the original name?

sherwinmina: Hi jspaine,

I removed the package-lock.json file. Used single quotes in the .js file. Also reset the input to its last state when canceled and edit is clicked again.

Issue #227, opened 2017-08-31
Restrict customer status updates for non-admins


jspaine:
The server customer controller update function allows customers to update their status even when currently rejected. It should only be allowed for customers that are active or inactive.

PR #239, opened 2017-09-16
customer controller middleware restricting customer status updates

davycheung:
Customer controller will restrict status changes for non-admins when updating a customer status from Reject or Pending.

Attempting to resolve Issue #227 .

### DIFF
--- a/server/controllers/customer.js
+++ b/server/controllers/customer.js
@@ -1,7 +1,7 @@
-import {extend, intersection} from 'lodash'
+import {extend, intersection, includes} from 'lodash'
 
 import {ForbiddenError, NotFoundError} from '../lib/errors'
-import {ADMIN_ROLE, clientRoles, volunteerRoles} from '../../common/constants'
+import {ADMIN_ROLE, clientRoles, volunteerRoles, customerStatus} from '../../common/constants'
 import Customer from '../models/customer'
 import mailer from '../lib/mail/mail-helpers'
 import User from '../models/user'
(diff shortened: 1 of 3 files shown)


server/routes/customer.js
```diff
@@ -12,6 +12,7 @@ const deleteSchema = {...sync, type: 'customer/DELETE_SUCCESS'}
 export default () => {
   const {requiresLogin} = userController
   const {hasAuthorization} = customerController
+  const {canChangeStatus} = customerController
```
jspaine: can just add this to the line above

server/controllers/customer.js
```diff
@@ -105,5 +105,29 @@ export default {
...
+      ADMIN_ROLE,
+    ]
+    const restrictedStatus = [
+      customerStatus.REJECTED,
+      customerStatus.PENDING,
+    ]
+
+    if (req.user && !intersection(req.user.roles, unrestrictedRoles).length) {
+      const customer = extend(req.customer, req.body)
+      const oldCustomer = await Customer.findById(customer._id)
```
jspaine: can't you just use `req.customer` then there's no need to look the customer up again?

server/routes/customer.js
davycheung: Updated in second commit.

server/controllers/customer.js
davycheung: Thanks, I've removed the second lookup in following commit.

jspaine:
closes #227

Issue #119, opened 2017-06-15
Need to have a generalized donation receipt that can be customized with the organization's name, EIN, contact info, etc.


jasonsreynolds:

jspaine:
Can you give an list of exactly what fields might be needed, or how variable is this stuff likely to be? Maybe it should be possible to add arbitrary fields to the general settings section.

jspaine:
I think to finish this there needs to be something like:
- [ ] An EIN field in the general settings
- [ ] A new placeholder for the receipt in `common/placeholders`. It should be required and added to the email during db seeding, like the password reset url placeholder. It should render a table with the donated items and values and total.
- [ ] Bonus if an example table can be rendered inside the editor too

- [ ] To make the app useful in other countries there could be a receipts section in the settings where the admin can add and remove arbitrary field name/value pairs. They should probably be rendered as part of the receipt table, maybe along with company name and some other fields from general settings, to avoid having to add too many placeholders to the menu.

ronaldblanco:
Hello Jspaine
Hey I am back, and I saw someone it is working on the issue 171.
I will be working on this.
In the "An EIN field in the general settings" do you only need the field with that name??

Thanks.

jspaine:
Hey, welcome back!

Yeah just that field is fine.

ronaldblanco:
Hey I did a pull request for the "An EIN field in the general settings" problem, this time I did check the eslint (server an client) to avoid future problems.
I will be checking the placeholder need (2 point) and I will ask you if I need your help.

Thanks.

ronaldblanco:
Hey jspaine;
In the second point; can I get the information for the table from a specific area of the app (existing process)? or must I get ed it from the database directly?

Thanks.

ronaldblanco:
Hey I need help;
I am trying to play as a donor (to test my changes) but before make anything I am unable to add my own donations (as a donor user); I get constantly error 401, unauthorized user on my console of the server.
Do I am missing something?? I had tried creating the user with the application and also writing directed to the database, but, same result all the time.  
Thanks.

davycheung:
Hi, @ronaldblanco I think I found the issue, and added a pull request to fix it.

ronaldblanco:
Thanks @davycheung, I will be testing it.

ronaldblanco:
Hello everyone;

I am working on the second point of this issue but I am unable to make my tests; what do I need to get my local app sending email so I can test my changes??
[...]

PR #241, opened 2017-09-21
EIN field in the general settings

ronaldblanco:
EIN field in the general settings was created to contributed to the solution of the issue #119.

The changes were tested locally and was also run the eslint check for client an server without any error.

### DIFF
--- a/client/modules/settings/components/General.js
+++ b/client/modules/settings/components/General.js
@@ -48,14 +48,6 @@ const General = () =>
 
       <Row>
         <Clearfix visibleSmBlock visibleMdBlock />
-        <Col lg={6}>
-          <RFFieldGroup
-            name="address"
-            label="Address"
-            type="textarea"
-            rows="5"
-          />
-        </Col>
         <Col lg={3}>
           <RFFieldGroup
             name="distanceUnit"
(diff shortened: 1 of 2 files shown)


client/modules/settings/components/General.js
```diff
@@ -68,14 +68,22 @@ const General = () =>
...
             type="text"
             required
           />
         </Col>
+        <Col lg={3}>
+          <RFFieldGroup
+            name="eIN"
+            label="EIN"
+            type="text"
+            required
```
kenjiO: I'm not sure how we are going to use EIN yet so can you delete the required prop here.  Or if you think it should be required can you add code for the server that validates EIN is present before saving to the database.  Thanks!

client/modules/settings/components/General.js
```diff
@@ -56,7 +56,7 @@ const General = () =>
             rows="5"
           />
         </Col>
-        <Col lg={3}>
+        <Col lg={1}>
```
jspaine: Isn't it a bit small and weird looking having 1 and 2 as the width of these? Why not keep them all 3 and move the address to the end if it's in the way?

client/modules/settings/components/General.js
```diff
@@ -68,14 +68,22 @@ const General = () =>
...
           <RFFieldGroup
             name="moneyUnit"
             label="Money Unit"
             type="text"
             required
           />
         </Col>
+        <Col lg={3}>
+          <RFFieldGroup
+            name="eIN"
```
jspaine: Why eIN and not EIN?

client/modules/settings/components/General.js
jspaine: It's for the donation receipt, but yeah, I think it shouldn't be required.

ronaldblanco:
Hello, I did follow the suggestions; take a look now.

Thanks.

jspaine:
Thanks, looks good.

PR #242, opened 2017-09-22
donation controller bug not allowing donors to create donations

davycheung:
Related to issue #119 - found a bug that prevents non-admin donors from creating their own donations.


jspaine:
Thanks!

PR #250, opened 2017-09-27
(#119-3)To make it look a bit more receipt like, table with EIN

ronaldblanco:
#119 -3 To make it look a bit more receipt like, there could be another table as part of the donation receipt with the organization details and EIN number from general settings.
Solution tested locally and checked with lint command.

### DIFF
--- a/common/placeholders.js
+++ b/common/placeholders.js
@@ -10,6 +10,7 @@ const placeholders = [
   {id: 'url', label: 'Foodbank Website'},
   {id: 'clientIntakeNumber', label: 'Client Intake Number'},
   {id: 'supportNumber', label: 'Support Number'},
+  {id: 'ein', label: 'Employer Identification Number'},
 
   {id: 'firstName', label: 'User First Name', type: placeholderTypes.EMAIL},
   {id: 'lastName', label: 'User Last Name', type: placeholderTypes.EMAIL},
@@ -34,11 +35,11 @@ const placeholders = [
     format: () => '<img src="cid:signature" alt="Signature" />'
   },
   {
-    id: 'items',
-    label: 'Donation Receipt Information',
+    id: 'receipt',
+    label: 'Donation Receipt',
     type: placeholderTypes.EMAIL,
     required: true,
-    format: value => table(value)
+    format: value => renderReceipt(value)
   },
 ]
 
@@ -66,14 +67,25 @@ function getPlaceholders(types) {
   return placeholders
 }
 
-function table(value) {
-  let tableFinal = `<table><tr><td><p>Item</p></td><td><p>Value</p></td></tr>`
-  let total = 0
-  for (let i = 0; i < value.length; i++) {
-    tableFinal += `<tr><td><p>"${value[i].name}"</p></td><td><p>"${value[i].value}"</p></td></tr>`
-    total += value[i].value
+function renderReceipt(values) {
+  const ids = ['organization', 'address', 'url', 'supportNumber', 'ein']
+  let table = '<table>'
+
+  for (let i = 0; i < ids.length; i++) {
+    const placeholder = placeholders.find(p => p.id === ids[i])
+    table += `<tr><td>${placeholder.label}:</td><td> ${values[ids[i]]}</td></tr>`
   }
-  return tableFinal + `<tr><td><p>Total:</p></td><td><p>"${total}"</p></td></tr></table>`
+
+  table += `<tr><td><p>Here you have a list of the donations our organization recibed from you:</p></td></tr>`
+  const items = values.items
+  table += `<tr><td><p>Item</p></td><td><p>Value</p></td></tr>`
+  let total = 0
+  for (let i = 0; i < items.length; i++) {
+    table += `<tr><td><p>${items[i].name}</p></td><td><p>${items[i].value}</p></td></tr>`
+    total += items[i].value
+  }
+
+  return table + `<tr><td><p>Total:</p></td><td><p>${total}</p></td></tr></table>`
 }
 
 export {
--- a/server/lib/mail/mail-generator.js
+++ b/server/lib/mail/mail-generator.js
@@ -3,7 +3,7 @@ import {readFileSync} from 'fs'
 import {resolve} from 'path'
 import striptags from 'striptags'
 import transformHtml, {h, getAttr, hasClass} from './html-transform'
-import {trim, uniq} from 'lodash'
+import {trim, uniq, isUndefined} from 'lodash'
 
 import config from '../../config'
 import Page from '../../models/page'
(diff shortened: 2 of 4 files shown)


jspaine:
I meant it should be part of the receipt placeholder. You can't edit tables with Quill and I'm not sure what'll happen if you insert one into the editor like that. What does it look like!?

ronaldblanco:
In the console.log looks well (the same way that the other)

jspaine:
I mean what does it look like in the email editor?

I think it should be in the same placeholder as the receipt.

ronaldblanco:
I will be checking.

ronaldblanco:
Hey @jspaine check now, I did create a new placeholder I think it is a better solution (as I understand)
Let me know.

Thanks.

common/placeholders.js
```diff
@@ -10,6 +10,7 @@ const placeholders = [
   {id: 'url', label: 'Foodbank Website'},
   {id: 'clientIntakeNumber', label: 'Client Intake Number'},
   {id: 'supportNumber', label: 'Support Number'},
+  {id: 'EIN', label: 'EIN Number'},
```
jspaine: Make the id lower case to match the others. Think the label should just be 'EIN' too, or 'Employer Identification Number'

server/models/settings.js
```diff
@@ -51,7 +51,7 @@ const SettingsSchema = new Schema({
   EIN: {
     type: String,
     trim: true,
-    default: ''
+    default: '0'
```
jspaine: I'd just leave this blank - `default: ''`

common/placeholders.js
```diff
@@ -76,6 +84,15 @@ function table(value) {
   return tableFinal + `<tr><td><p>Total:</p></td><td><p>"${total}"</p></td></tr></table>`
 }
 
+function tableReceipt(placeholders) {
+  let head = ['Foodbank Name','Foodbank Address','Foodbank Website','Client Intake Number', 'Support Number','EIN Number']
+  let tableFinal = `<table>`
+  for (let i = 0; i < 6; i++) {
+    tableFinal += `<tr><td><p>"${head[i]}":</p></td><td><p>"${placeholders[i]}"</p></td></tr>`
```
jspaine: Think a better way to do this is something like:
```js
function renderReceipt(values) {
  const ids = ['organization', 'address', 'url', 'supportNumber', 'ein']
  let table = '<table>'

  for (let i = 0; i < ids.length; i++) {
    const placeholder = placeholders.find(p => p.id === ids[i])
    table += `<tr><td>${placeholder.label}: ${values[placeholder.id]}</td></tr>`
  }

  // then generate the items table
}
```
The `server/lib/mail/mail-helpers.js` `send` function already passes the settings fields so you shouldn't need to change the `bindPlaceholders` function in the mail generator.

server/models/settings.js
```diff
@@ -51,7 +51,7 @@ const SettingsSchema = new Schema({
   EIN: {
```
jspaine: Should also be lower case

ronaldblanco:
Hello @jspaine; check now, I did left the ein default: '0' because the ein placeholder have problems with an empty string.

server/lib/mail/mail-generator.js
```diff
@@ -74,9 +74,18 @@ function transformSubject(subject, bindings) {
 }
 
 function bindPlaceholder(node, bindings, attachments) {
+
   const id = getAttr(node, 'data-id')
   const placeholder = placeholders.find(pl => pl.id === id)
 
+  //Receipt Placeholder only
+  if (id === 'receipt') {
```
jspaine, line 82: Could probably just do:
```js
if (id === 'receipt') {
  bindings.receipt = {...bindings}
}
```

common/placeholders.js
```diff
@@ -40,6 +41,13 @@ const placeholders = [
     required: true,
```
jspaine: remember to delete this placeholder and the `table` function

ronaldblanco: You do not think it could be use-full later??

server/lib/mail/mail-generator.js
ronaldblanco: Yes, later I did think on that.

ronaldblanco:
Check now and let me know!!

server/lib/mail/mail-generator.js
jspaine: That's ok too, but you can just return `h(placeholder.format(bindings))` i think. `format` should be a function, and if it's not, it doesn't make much sense to return a text node of the bindings object anyway.

common/placeholders.js
```diff
@@ -66,14 +67,25 @@ function getPlaceholders(types) {
...
+
+  table += `<tr><td><p>Here you have a list of the donations our organization recibed from you:</p></td></tr>`
+  const items = values.items
+  table += `<tr><td><p>Item</p></td><td><p>Value</p></td></tr>`
   let total = 0
-  for (let i = 0; i < value.length; i++) {
-    tableFinal += `<tr><td><p>"${value[i].name}"</p></td><td><p>"${value[i].value}"</p></td></tr>`
-    total += value[i].value
+  for (let i = 0; i < items.length; i++) {
+    table += `<tr><td><p>"${items[i].name}"</p></td><td><p>"${items[i].value}"</p></td></tr>`
```
jspaine: Why the " 's around item name, value and total? Doesn't it look weird?

common/placeholders.js
jspaine: I don't see how. In any case, if you're not actually planning to use it for something, get rid of it. It'll still be in the git history.

jspaine:
What was the problem with the ein field being an empty string by default?

ronaldblanco:
When ein = "" the placeholder ein does not return, it give a problem of "No placeholder with that id" or something like that.
I thing that have to be with the mail-generator.js, some validation missing.

ronaldblanco:
OK, I will be checking the rest you told me.

jspaine:
Ah I think it's line 89 in the mail generator:
```js
if (!bindings[id] && placeholder.type !== placeholderTypes.ATTACHMENT)
```
should probably be:
```js
// import {isUndefined} from 'lodash'
if (isUndefined(bindings[id]) && placeholder.type !== placeholderTypes.ATTACHMENT)
```

common/placeholders.js
ronaldblanco: What are you referring exactly??
I do not understand!

ronaldblanco:
You were right with the correction on line 89.
Check now and let me know.

common/placeholders.js
jspaine: Sorry! I don't think you need the double quotes, it should just show $15 for example, not "$15".

ronaldblanco:
Ready.

jspaine:
Thanks!

Issue #255, opened 2017-10-01
Webpack fails to compile on admin-lte 2.4.0


tombarton:
Webpack fails to compile with the following errors when using admin-lte 2.4.0:

```
ERROR in ./client/app.js
Module not found: Error: Can't resolve 'admin-lte/plugins/fastclick/fastclick' in '/Users/tombarton/Sites/pantry-for-good/client'
 @ ./client/app.js 39:0-48
 @ multi react-hot-loader/patch webpack-dev-server/client?http://localhost:8080 webpack/hot/only-dev-server whatwg-fetch ./client/app.js

ERROR in ./client/app.js
Module not found: Error: Can't resolve 'admin-lte/plugins/slimScroll/jquery.slimscroll' in '/Users/tombarton/Sites/pantry-for-good/client'
 @ ./client/app.js 37:0-57
 @ multi react-hot-loader/patch webpack-dev-server/client?http://localhost:8080 webpack/hot/only-dev-server whatwg-fetch ./client/app.js
```

Issue can be resolved by downgrading to admin-lte 2.3.11. Happy to submit a PR if required.

jspaine:
Yeah it would be great if you could update it and fix that.

PR #256, opened 2017-10-01
Fix `admin-lte` version issue

tombarton:
Force use of `admin-lte` ~2.3, as 2.4.0 is currently breaking the webpack compile.

Addresses #255.


jspaine:
Thanks!

PR #259, opened 2017-10-03
Invert orientation of the pagination dropdown menu

Knochenmark:
Instead of showing the dropdown menu orientated to the bottom of the button, invert it and show it above, to prevent the dropdown menu to overflow out of the bottom of the page. This happened to me quite often, when the page wasn't scrolled down completely and the pagination button was just barely visible. When clicking it I was only able to see like the "10" and "20" but often the "30" and "50" were not visible at all.

Not entirely sure this is in the right place, but I couldn't really find any specific styles for this except the default bootstrap styles. Feel free to guide me to the right place.

### DIFF
--- a/client/modules/food/components/inventory/FoodItems.js
+++ b/client/modules/food/components/inventory/FoodItems.js
@@ -223,6 +223,13 @@ class FoodItems extends React.Component {
     }
   }
 
+  renderSizePerPageDropDown = (props) => {
+    return (
+      <SizePerPageDropDown
+        variation='dropup'/>
+    );
+  }
+
   render = () => {
     // set options for react-bootstrap-table
     const tableOptions = {
@@ -230,6 +237,7 @@ class FoodItems extends React.Component {
       toolBar: this.createCustomToolBar,
       defaultSortName: 'name',
       defaultSortOrder: 'asc',
+      sizePerPageDropDown: this.renderSizePerPageDropDown,
       noDataText: (this.props.foodCategories.length === 0)
         ? 'No foods in inventory. Add a category prior to adding a food'
         : 'No foods in inventory matching ' + this.state.searchText,
--- a/client/modules/food/components/packing/Packages.js
+++ b/client/modules/food/components/packing/Packages.js
@@ -35,7 +35,14 @@ class Packages extends Component {
   formatContents = contentList => {
     return contentList.reduce((prev, curr) => {return `${prev} ${curr.name},`}, "")
   }
-  
+
+  renderSizePerPageDropDown = (props) => {
+    return (
+      <SizePerPageDropDown
+        variation='dropup'/>
+    );
+  }
+
   getActionButtons = (_, foodPackage) =>
     <div>
       <Button bsStyle="danger"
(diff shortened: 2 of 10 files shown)


jspaine:
You can apparently [customize the dropdown](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/custom/pagination/custom-size-per-page.js#L39). Let me know if you try it, or I will when I get the chance.

kenjiO:
The documentation for using the customized drop down is [http://allenfang.github.io/react-bootstrap-table/custom.html#sizePerPageDropdown](http://allenfang.github.io/react-bootstrap-table/custom.html#sizePerPageDropdown)

Knochenmark:
I will probably look into it tonight, which means GMT+1 tonight. Looking at your example it might be just adding "dropup" instead of "dropdown" then?

Knochenmark:
So far i fixed the following lists, there are two i have some issues with, maybe someone can guide me where i find these two bootstrap tables?

- [x] Customer
- [x] Volunteers
- [x] Donors
- [ ] Food Schedule
- [x] Packing
- [ ] Inventory
- [x] Drivers
- [x] User Accounts

client/modules/food/components/packing/Packages.js
```diff
@@ -55,6 +62,7 @@ class Packages extends Component {
             options={{
               defaultSortName: "datePacked",
               defaultSortOrder: 'desc',
+              sizePerPageDropDown: this.renderSizePerPageDropDown,
```
Knochenmark, line 65: this seems to be ignored for some reason. Any idea why?

client/modules/food/components/inventory/FoodItems.js
```diff
@@ -223,13 +223,21 @@ class FoodItems extends React.Component {
...
+  }
+
   render = () => {
     // set options for react-bootstrap-table
     const tableOptions = {
       // toolBar specifies the function to create the table header
       toolBar: this.createCustomToolBar,
       defaultSortName: 'name',
       defaultSortOrder: 'asc',
+      sizePerPageDropDown: this.renderSizePerPageDropDown,
```
Knochenmark, line 240: this also seems like its not applied.

Knochenmark:
I've amended the changes and added comments to the questionable files. Any idea why these two don't work?

kenjiO:
I tried your Packages.js and FoodItems.js changes and they both worked for me in Chrome 61 on a mac.  What browser and OS were you getting problems in?

Also I would suggest that instead of defining the renderSizePerPageDropDown function to just use an anonymous function since it is a one liner.
```javascript
<BootstrapTable
    options={{
        sizePerPageDropDown: () => <SizePerPageDropDown variation='dropup'/>
    }}   
>
```

jspaine:
Just tried it in firefox and chrome, they all seem to work.

kenjiO:
Ah yes, I remember reading about not using the inline functions now.  Good call.

Knochenmark:
Oh, it works for you? I will try it again tonight, maybe the hotswap just didnt work for me. I will verify this. I am on Chrome Osx.

Knochenmark:
Ok, i just checked and seems like all lists have an inverted dropdown now. It must have been some issue with the hotswap feature on my side.

jspaine:
Nice, thanks for the fix.

PR #260, opened 2017-10-03
Fix/packing controller check quality

### LINKED ISSUE #248: Packing Controller Needs to Check Quantity
Before creating a pack the controller needs to check that all items selected for the pack have sufficient quantities. 

screen shot 2017-09-26 at 7 59 27 pm

ik226:
client/modules/food/components/inventory/FoodItems.js:125 ~138

When `this.state.hasBeenChanged` stays `false` (default), 
this line 
```javascript
(this.state.modalInputFields.quantity !== "" && this.state.modalInputFields.quantity >= 0) || !this.state.hasBeenChanged  ? null : 'error'
```
 allows user to put any quantity value, including negative values. Also, I believe that `this.state.hasBeenChanged` will only become `true` when user changes food name. 

Thus, the approach is to remove parts that checking `this.state.hasBeenChanged`. Instead, we can 
store initial values (name, category, quantity) for each food item when an edit-modal appears. Then, we can decide to disable update button based on comparison between initial values and new inputs.

closes #248

### DIFF
--- a/client/modules/food/components/inventory/FoodItems.js
+++ b/client/modules/food/components/inventory/FoodItems.js
@@ -20,7 +20,6 @@ class FoodItems extends React.Component {
       modalInputFields: { name: "", categoryId: "", quantity: "" },
       validInput: false,
       searchText: "",
-      hasBeenChanged: false,
 
       //store initial value
       initialModalInputFields: { name: "", categoryId: "", quantity: "" },
@@ -44,25 +43,20 @@ class FoodItems extends React.Component {
         showModal: 'Add',
         editModalFood: undefined,
         modalInputFields: { name: "", categoryId: "", quantity: "" },
-        hasBeenChanged: false
       })
     } else {
       // Open the modal in 'edit' mode
       const food = this.props.foodItems.find(food => food._id === _id)
+      const inputFields = {
+        name: food.name,
+        categoryId: food.categoryId,
+        quantity: food.quantity.toString()
+      }
       this.setState({
         showModal: 'Edit',
         editModalFood: food,
-        initialModalInputFields: {
-          name: food.name,
-          categoryId: food.categoryId,
-          quantity: food.quantity.toString()
-        },
-        modalInputFields: {
-          name: food.name,
-          categoryId: food.categoryId,
-          quantity: food.quantity
-        },
-        hasBeenChanged: false
+        initialModalInputFields: inputFields,
+        modalInputFields: inputFields,
       })
     }
     this.props.clearFlags()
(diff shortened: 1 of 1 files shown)


client/modules/food/components/inventory/FoodItems.js
```diff
@@ -20,7 +20,10 @@ class FoodItems extends React.Component {
       modalInputFields: { name: "", categoryId: "", quantity: "" },
       validInput: false,
       searchText: "",
-      hasBeenChanged: false
+      hasBeenChanged: false,
```
jspaine: It doesn't look like this is still used after this change

client/modules/food/components/inventory/FoodItems.js
```diff
@@ -49,6 +52,11 @@ class FoodItems extends React.Component {
       this.setState({
         showModal: 'Edit',
         editModalFood: food,
+        initialModalInputFields: {
```
jspaine: Could just define a variable for setting modalInputFields and initialModalInputFields so you don't have to write it twice

client/modules/food/components/inventory/FoodItems.js
```diff
@@ -145,6 +154,14 @@ class FoodItems extends React.Component {
   }
 
   /**
+   *  check whether newly edited values are different from the inital values
+   */
+   checkChanged = () => {
+     //console.log(JSON.stringify(this.state.initialModalInputFields), JSON.stringify(this.state.modalInputFields))
```
jspaine: Remeber to remove

client/modules/food/components/inventory/FoodItems.js
```diff
@@ -145,6 +154,14 @@ class FoodItems extends React.Component {
...
+     return JSON.stringify(this.state.initialModalInputFields) === JSON.stringify(this.state.modalInputFields) ? false : true
```
jspaine: Rather check explicitly, something like
```js
mapKeys(    // from lodash
  this.state.modalInputFields, 
  (value, key) => value !== this.state.initialModalInputFields[key]
).reduce((acc, x) => acc || x, false)
```

client/modules/food/components/inventory/FoodItems.js
ik226: can you please elaborate this part further?

jspaine: just like this, so if you add/remove a field you don't have to change it in two places:
```js
const inputFields = {...}
this.setState({
   ...
   initialModalInputFields: inputFields,
   modalInputFields: inputFields
})
```

ik226: I see! Thanks so much!

jspaine:
Thanks

PR #261, opened 2017-10-05
(#119-4) Instead of having the EIN field hard coded there could be a settings …

ronaldblanco:
Related to #119 point 4: Instead of having the EIN field hard coded there could be a settings section for extra receipt related fields, similar to the create donation items fields and household section on the customer application. So by default it would be one field, Name: EIN, Value: whatever, but you could rename it or add other fields appropriate for your country.

Create a receiptFields-schema.js file to be used with the settings.js model.
All changes were tested locally and with lint command.

### DIFF
--- a/client/modules/settings/components/General.js
+++ b/client/modules/settings/components/General.js
@@ -68,13 +68,6 @@ const General = () =>
             required
           />
         </Col>
-        <Col lg={3}>
-          <RFFieldGroup
-            name="EIN"
-            label="EIN"
-            type="text"
-          />
-        </Col>
         <Clearfix visibleSmBlock visibleMdBlock />
         <Col lg={6}>
           <RFFieldGroup
--- a/client/modules/settings/components/ReceiptFields.js
+++ b/client/modules/settings/components/ReceiptFields.js
@@ -1,0 +1,81 @@
+import React from 'react'
+import {Clearfix, Button} from 'react-bootstrap'
+
+import {Box, BoxHeader} from '../../../components/box'
+import {RFFieldGroup} from '../../../components/form'
+
+import { FieldArray } from 'redux-form'
+
+const renderRow = ({ fields, meta: { error, submitFailed } }) => (
+
+  <div style={{padding: '10px'}}>
+    <ul style={{paddingLeft: 0}}>
+      <h4>Receipt Fields:</h4>
+      {fields.map((receiptField, index) => (
+        <li key={index} style={{display: 'flex', flexWrap: 'nowrap'}}>
+          <div style={{display: 'flex', flexGrow: 1}}>
+            <RFFieldGroup
+              name={`${receiptField}.name`}
+              type="text"
+              placeholder="Name"
+              style={{
+                flexGrow: 1,
+                margin: '10px 10px 0 0'
+              }}
+            />
+            <RFFieldGroup
+              name={`${receiptField}.value`}
+              type="text"
+              placeholder="Value"
+              style={{
+                flexGrow: 1,
+                margin: '10px 0px 0 0'
+              }}
+            />
+            <Button
+              type="button"
+              title="Remove ReceiptField"
+              onClick={() => fields.remove(index)}
+              className="btn"
+              style={{
+                margin: '10px 0 0px 18px',
+                height: '34px'
+              }}
+            >
+              <i className="fa fa-trash" />
+            </Button>
+            <Clearfix visibleSmBlock visibleMdBlock />
+          </div>
+        </li>
+      ))}
+      <Button
+        type="button"
+        onClick={() => fields.push({})}
+        className="btn"
+        style={{
+          margin: '15px 0px 0px 15px',
+          height: '34px'
+        }}
+      >
+        Add Field
+      </Button>
+      {submitFailed && error && <span>{error}</span>}
+    </ul>
+  </div>
+)
+
+const ReceiptFields = props => {
+  const { handleSubmit } = props
+  return (
+    <Box>
+      <BoxHeader heading="General Receipt Fields" />
+      <form onSubmit={handleSubmit} >
+        <div className="form-group">
+          <FieldArray name="receiptFields" component={renderRow} />
+        </div>
+      </form>
+    </Box>
+  )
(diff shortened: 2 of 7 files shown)


ronaldblanco:
Hello @jspaine, take a look at this to see if that's the idea.
As I understand the ein field will be part now of this new schema right??

jspaine:
Yeah, you can just add it to the settings model file though because it's not used anywhere else.

ronaldblanco:
Hey @jspaine  take a look now.

Thanks.

ronaldblanco:
Take a look also to the place holder modification.

server/controllers/settings.js
```diff
@@ -35,12 +35,14 @@ export default {
...
     const location = await locateAddress(req.body.address)
+    const receiptFields = req.body.receiptFields
 
     if (!location) return next(new ValidationError({address: 'Address not found'}))
 
     const settings = {
       ...req.body,
-      location
+      location,
+      receiptFields
```
jspaine: The `...req.body` above will also copy this, you shouldn't need either of the changes in this file.

client/modules/settings/components/General.js
```diff
@@ -70,8 +70,15 @@ const General = () =>
         </Col>
         <Col lg={3}>
           <RFFieldGroup
-            name="EIN"
-            label="EIN"
+            name="receiptFields.name"
```
jspaine: The idea was you'd be able to add/remove multiple of these fields using a [redux-form FieldArray](https://redux-form.com/6.0.0-alpha.7/examples/fieldarrays/). There are other examples of its use in

> create donation items fields and household section on the customer application.

ronaldblanco: Ok, I am working on this, but I thing I it would take me a little of time.

Can you also suggest me a good simple example or starter guia about the Redux-React combination? I need to check this as an first time user because I have not get this concept yet.

Thanks.

jspaine: Yeah that's ok, there's a lot to learn around that though.

I thought this (and the second part) was great https://egghead.io/courses/getting-started-with-redux

And for functional programming I think the first parts of this are pretty nice http://reactivex.io/learnrx/

ronaldblanco:
Hello @jspaine I am back of my research and study.
Hey check this last commit, I think it fulfill the requirement; if it is so I will them rebase the branch in order to give you only 1 completed commit.

Thanks.

jspaine:
Hi! It looks about right, I'll have a closer look later. The layout is a bit off though (try looking at the styling of the other similar components) and I don't think we need a title (Receipt Field #1), or even labels (Name, Value) for each row.

Also calling it RF is a bit confusing in the code because there are a few redux-form based components abbreviated the same, and it's not going to be helpful for users to see it in the ui. Just call it ReceiptFields or something in the code and have 'Add Field' or 'Add Row' in the ui.

ronaldblanco:
OK, Here I fixed your suggestions.
Let me know.

client/modules/settings/components/ReceiptFields.js
```diff
@@ -0,0 +1,96 @@
+import React from 'react'
+import {Clearfix, Col, Row, Button} from 'react-bootstrap'
+
+import {Box, BoxBody, BoxHeader} from '../../../components/box'
+import {RFFieldGroup} from '../../../components/form'
+
+import { FieldArray } from 'redux-form'
+
+const renderField = ({ input, label, type, meta: { touched, error } }) => (
```
jspaine: I don't think this is needed, RFFieldGroup already renders an input so you can just remove this function and the `component` props from the field groups below and it still works.

client/modules/settings/components/ReceiptFields.js
```diff
@@ -0,0 +1,96 @@
...
+  <div>
+    <label>{label}</label>
+    <div>
+      <input {...input} type={type} placeholder={label} className="form-control" />
+      {touched && error && <span>{error}</span>}
+    </div>
+  </div>
+)
+
+const renderRF = ({ fields, meta: { error, submitFailed } }) => (
```
jspaine: I'd just call this `row` or `renderRow`

client/modules/settings/components/ReceiptFields.js
```diff
@@ -0,0 +1,96 @@
...
+        }}
+      >
+        Add Field
+      </Button>
+      {submitFailed && error && <span>{error}</span>}
+    </Row>
+  </BoxBody>
+)
+
+const GeneralRF = props => {
```
jspaine: And call this `ReceiptFields` like the filename

server/config/mailer.js
```diff
@@ -49,6 +49,6 @@ export default async function sendEmail(
     path: '/v3/mail/send',
     body: mail.toJSON()
   })
-
+  //console.log(content)
```
jspaine: Remember to clean this up

server/controllers/settings.js
```diff
@@ -35,12 +35,13 @@ export default {
     }
 
     const location = await locateAddress(req.body.address)
+    // const receiptFields = req.body.receiptFields
```
jspaine: Also clean up changes in this file

ronaldblanco:
OK, check now!
It is already squash in one commit only.

jspaine:
Looks good, the columns are still `<Col lg={3}>` though, so at small and medium breakpoints they take up the full width. You could try using `sm={5}` for the two field columns and `sm={2}` for the button column (it should add up to 12 for the full width), but the button column will look too big in wide browser windows.

I think I used flexbox on the customer household form, set the buttons as fixed width and let the fields grow so they share the remaining space.

ronaldblanco:
Take a look!

jspaine:
It's better, looks a bit weird with no space between name and value fields though.

ronaldblanco:
Sorry I do not understand; I see it like the Household example.

jspaine:
The household one has margin:10px on the inputs, so there's a gap between them.

ronaldblanco:
I get it; check now!

jspaine:
Looks good! Thanks!

PR #263, opened 2017-10-06
add .travis.yml for Travis CI service

kenjiO:


jspaine:
What were you thinking to do with this? Currently tests run locally on commit, but I could maybe see a use in verifying the production build. Or for running eg. functional tests that might take a while locally.

kenjiO:
I think CI is a common practice now and I think it would be good to have tests run on a different platform in case anything specific to a local machine is causing a test to pass.  Also I was thinking of using https://coveralls.io/ for code coverage which requires travis.

PR #264, opened 2017-10-08
strict selection of markers when click MarkerCluster

### LINKED ISSUE #215: Driver assignment marker cluster selection
Clicking a marker cluster on the map selects some customers from adjacent clusters.

ik226:
- client/modules/driver/reducers/assignment.js

Apparently, MarkerClusterer’s bound is ‘larger’ than it is supposed to. As a result,  `MarkerCluster.isMarkerInClusterBounds()` returns true to those markers located nearby but not inside the cluster.
So, how about we instead use `MarkerClusterer.getMarkers()`, which returns array of the markers assigned to this cluster. Then we can compare their location coordinates (lat,lng) with those of customer markers, and return those match each other. 

My concern is the way the code compares customer makers to markers in each cluster. It looks for matching lat/lng coordinates by ‘assuming’ all customer makers have coordinates with 6 decimal points. (this works just fine with current data set)

- client/modules/driver/reducers/assignment.spec.js 

Above modification requires our mock cluster object to have method getMarkers(). 

Please let me know if have any suggestions!

closes #215

### DIFF
--- a/client/modules/driver/components/driver-assignment/map/Markers.js
+++ b/client/modules/driver/components/driver-assignment/map/Markers.js
@@ -71,7 +71,6 @@ const Markers = ({
           icon={isSelected(customer._id) ? selectedCustomerIcon : customerIcon}
           position={customer.location}
           onClick={toggleCustomer(customer._id)}
-          title={customer._id.toString()}
         />
       )}
     </MarkerClusterer>
--- a/client/modules/driver/components/driver-assignment/map/Markers.js
+++ b/client/modules/driver/components/driver-assignment/map/Markers.js
@@ -71,6 +71,7 @@ const Markers = ({
           icon={isSelected(customer._id) ? selectedCustomerIcon : customerIcon}
           position={customer.location}
           onClick={toggleCustomer(customer._id)}
+          title={customer._id.toString()}
         />
       )}
     </MarkerClusterer>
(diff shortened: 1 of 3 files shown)


jspaine:
Yeah it's probably not very robust to assume the marker and corresponding customer coordinates will be exactly the same. It seems to work to manually create the bounds with `google.maps.LatLngBounds(southWest, northEast)` then use its `contains` method to filter the customers.

ik226:
I see. This problem could be easily solved if we can assign customer ids inside a custom property of each marker.  If this is possible, we can eliminate extra iterations over whole customer lists and bound calculation.
Hopefully, something like below: 
```javascript
{customers && customers.filter(c => c.location).map(customer =>
        <Marker
          key={customer._id}
          icon={isSelected(customer._id) ? selectedCustomerIcon : customerIcon}
          position={customer.location}
          onClick={toggleCustomer(customer._id)}
          //customerId={customer._id} //** something like this
        />
      )}
```
However, react-google-maps does not allow passing any arbitrary property when instantiate Marker object.

But, we can be little creative: how about we assign corresponding customer IDs to each marker's 'title' property. We can achieve two advantages (as I believe): 
- it can efficiently get customer ids by doing `MarkerCluster.getMarkers().map(m => m.title)` without implementing bound calculations.
- Also, since title property is for showing title inside tooltip to each marker when rolled over, we can improve UX as each marker can easily be identified with their id number shown inside one's tooltip. This can help usability of map interaction.

ik226:
(oops.. mistakenly closed PR..)

jspaine:
I think it's bad practice to tie the calculation together with the ui like that. 

Even with thousands of markers and customers, calculating bounds and filtering is still going to be really quick. It's up to 4 iterations over the markers in the cluster (could be optimized to 1 if necessary) and one over the customers.

jspaine:
Nice! Thanks!

PR #265, opened 2017-10-08
added default frequency to 1 and start date to current date to food m…

icpmacdo:
…odel

### DIFF
--- a/server/models/food.js
+++ b/server/models/food.js
@@ -1,42 +1,43 @@
 import mongoose from 'mongoose'
 
-import {modelTypes} from '../../common/constants'
+import { modelTypes } from '../../common/constants'
 
-const {Schema} = mongoose
+const { Schema } = mongoose
 
 const FoodItemSchema = new Schema({
-  name: {
-    type: String,
-    trim: true
-  },
-  quantity: {
-    type: Number,
-    default: 0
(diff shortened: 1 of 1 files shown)


server/models/food.js
```diff
@@ -14,11 +14,11 @@ const FoodItemSchema = new Schema({
     default: 0
   },
   startDate: {
-    type: Date
+    type: Date.now
```
jspaine: `type` should still be `Date`, just the default should be `Date.now`

server/models/food.js
```diff
@@ -1,43 +1,44 @@
...
-  },
-  frequency: {
-    type: Number,
-    default: 0
-  },
-  deleted: {
-    type: Boolean,
-    default: false
-  }
+    name: {
```
jspaine: Looks good, but now the indentation's wrong. You might have to set your editor to use 2 spaces for tabs.

icpmacdo:
Hey, I think the indentation is not correct. Need to figure out why VS Code wants to screw it up.

jspaine:
Cool, thanks. It's just the default I think, you can add `"editor.tabSize": 2,` to your user settings (or workspace settings if you only want it like that for this).

PR #266, opened 2017-10-12
properly normalize state data in api.spec.js getLinkableFields test

kenjiO:
The state in this test did not have normalized data in it.  For some reason normalizr versions before 3.2.4 didn't have a problem with it so the tests have been passing up until recently when 3.2.4 was released.


jspaine:
Ah, good job figuring that out!

PR #270, opened 2017-10-18
Feature/food inventory touched

Thirdoptics:
Addresses  
> Should show errors only on touched fields, [Food Inventory #171](https://github.com/freeCodeCamp/pantry-for-good/issues/171)

added touched state to: 
 - constructor
 - openModal
 - closeModal

added touch functions to be triggered by inputs onFocus

amended validation logic by adding !this.state.touched.fieldName || (existing logic )

### DIFF
--- a/client/modules/food/components/inventory/FoodItems.js
+++ b/client/modules/food/components/inventory/FoodItems.js
@@ -20,7 +20,8 @@ class FoodItems extends React.Component {
       modalInputFields: { name: "", categoryId: "", quantity: "" },
       validInput: false,
       searchText: "",
-
+      //touched input fields
+      touched: { foodName: false, foodCategory: false, foodQuantity: false },
       //store initial value
       initialModalInputFields: { name: "", categoryId: "", quantity: "" },
     }
@@ -43,6 +44,7 @@ class FoodItems extends React.Component {
         showModal: 'Add',
         editModalFood: undefined,
         modalInputFields: { name: "", categoryId: "", quantity: "" },
+        touched: { foodName: false, foodCategory: false, foodQuantity: false },
       })
     } else {
       // Open the modal in 'edit' mode
@@ -57,6 +59,7 @@ class FoodItems extends React.Component {
         editModalFood: food,
         initialModalInputFields: inputFields,
         modalInputFields: inputFields,
+        touched: { foodName: true, foodCategory: true, foodQuantity: true },
       })
     }
     this.props.clearFlags()
@@ -68,6 +71,7 @@ class FoodItems extends React.Component {
       showModal: false,
       editModalFood: undefined,
       modalInputFields: { name: "", categoryId: "", quantity: "" },
+      touched: { foodName: false, foodCategory: false, foodQuantity: false },
       validInput: false,
     })
     this.props.clearFlags()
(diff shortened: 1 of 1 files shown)


client/modules/food/components/inventory/FoodItems.js
```diff
@@ -146,6 +150,25 @@ class FoodItems extends React.Component {
...
+    foodName: () => this.setState({ touched: { ...this.state.touched, foodName: true } }, this.validate),
+    foodCategory: () => this.setState({ touched: { ...this.state.touched, foodCategory: true } }, this.validate),
+    foodQuantity: () => this.setState({ touched: { ...this.state.touched, foodQuantity: true } }, this.validate),
+  }
+
+  /**
+   *  callback for setState when user selects an existing food name from Autosuggest
+   */
+  validateFocus = () => {
+    this.validate
```
jspaine: Did you mean to call this?

jspaine: Did you mean to call this? It seems to behave the same either way.

Maybe something like
```js
if (this.validate()) {
  this.quantity.focus()
  ...
```

Thirdoptics:
I thought I'd done something a little wonky with the branches/pull requests, thanks for pointing that out, getting a better picture of the ins and outs of git now.  Would you like me to go back and split it to 2 pull requests or just leave it for this?  

I'll update the code with the suggested if(validate), that's a subtle but good change, no reason to point to focus to an invalid field.  

> This file's getting a bit complicated/large, and handling form state manually is a bit difficult too.
I agree, but it does help with understanding things like redux form.  I was wondering why redux form isn't in use here?

jspaine:
It's fine, I'll just squash and merge both so it doesn't make any difference. :smile: 

I think just that it was simple enough that redux form didn't seem necessary.

Thirdoptics:
Thanks, I'll get this whole git thing figured out sooner or later.  Had the same thoughts on redux form, just curious if my logic was in the right direction. 

Made the updates to validateFocus, also added a conditional to move the cursor only when editing inventory.

```
if(this.state.showModal === 'Edit'){
  let storedValue = this.quantity.value
  this.quantity.value = ''
  this.quantity.value = storedValue
}
```

client/modules/food/components/inventory/FoodItems.js
```diff
@@ -146,6 +150,27 @@ class FoodItems extends React.Component {
...
+    foodName: () => this.setState({ touched: { ...this.state.touched, foodName: true } }, this.validate),
+    foodCategory: () => this.setState({ touched: { ...this.state.touched, foodCategory: true } }, this.validate),
+    foodQuantity: () => this.setState({ touched: { ...this.state.touched, foodQuantity: true } }, this.validate),
+  }
+
+  /**
+   *  callback for setState when user selects an existing food name from Autosuggest
+   */
+   validateFocus = () => {
+     if(this.validate){
```
jspaine: `this.validate` is a function though, so it'll always be truthy. Looking at it more, `validateFocus` is only called after setting state to an existing food item, so it'll always be valid anyway and checking isn't necessary. 

It's a bit weird if you type a complete existing food name, the dropdown stays open and focus moves to quantity. The `closeOnCompletion` prop doesn't seem to help.

Thirdoptics: Ahh, good catch on this.Validate always being truthy, can't believe I missed the point of that 🤦 .  

I'll take another stab at this tonight/tomorrow and see if I can get the weird behavior figured out.

jspaine: Cool, I don't know if it's worth spending much time on though, seems a bit of an edge case.

Thirdoptics: Got it cleaned up a bit, removed the ValidateFocus since it's largely unnecessary, the only thing lost is moving the cursor for the case where someone edits an item and uses autosuggest to select a new item.

I think the weird behavior with Autosuggest and the `closeOnCompletion` is an internal problem with Autosuggest, I couldn't get that prop to work even outside this project in a simple setting.

I did find a react warning when you open the autosuggest list when more than one item has the same name.  It can be fixed by changing `itemReactKeyPropName` to id instead of name.  Wasn't sure if I should open an issues for something so trivial or just fix it here?

> Warning: flattenChildren(...): Encountered two children with the same key, `.0:$Peas`. Child keys must be unique; when two children share a key, only the first child will be used.

jspaine:
Thanks!

jspaine:
Nice! I'm going to merge it but want to merge both so you get two contributions, but I don't think it'll be as easy as I first thought. I'll do it soon though.

Thirdoptics:
Thanks for all the help @jspaine, learned a lot through the process.  Double thanks for the extra work to give credit for both contributions 😄

PR #282, opened 2017-10-25
fixes #281 shows address as required in general settings

### LINKED ISSUE #281: Add * to address field label on settings
In the settings section, general page, the address field label should have a * to indicate it is a required field.

logangingerich:
Fix for issue #281, now address (under settings/general) has an '*' to show that it is a required field.


kenjiO:
Looks great.  Thanks for the fix.

PR #285, opened 2017-10-26
Organization name cannot be only empty spaces

### LINKED ISSUE #280: Settings should not allow blank Organization name
In the settings section, general page, it accepts just spaces as a valid entry.  It should make sure there is at least one non-space character for the organization name.

almozaz:
Closes #280


kenjiO:
Nice fix! Thanks.

Issue #130, opened 2017-06-15
Tests


jspaine:

dmcfaddengalway:
Hey kenjio, I'd love to help out with this. I've been learning about unit tests recently and want to see what I can apply. Let me know if this is free to try.

jspaine:
Hi, there's quite a lot left to test so just do some if you want. The chances of someone else submitting tests for the same thing at the same time are pretty slim I think.

Thirunavukkarasu:
@jspaine I'm interested in writing test cases for some the remaining modules. Please let me know for which i'm allowed try.

Thanks!

kenjiO:
You are welcome to test anything you find that you think needs more testing.  You can run the `npm run coverage` task an it will generate a test report in **coverage/lcov-report/index.html** that shows which parts of the codes the tests are touching.    If you find something you want to test I would just open up an issue for the specific files you are testing so no one else works on it.

Thirunavukkarasu:
@kenjiO Kenji - I will writing tests for questionnaire-view and section-view.

QuestionnaireView.js
SectionView.js

Thanks.

vanessasena:
Hi, when I execute `npm run test` I got the error "Fatal RangeError: Maximum call stack size exceeded". Does anyone know what is this?

jspaine:
Hi, can you post the stacktrace, and whether it happens in client or server tests? (`npm run test:client` and `npm run test:server`)

vanessasena:
Hi @jspaine the problem is in the server, in this [test](https://github.com/freeCodeCamp/pantry-for-good/blob/a8fd63e509389f516cbb8d556c61cb03e221e63f/server/tests/integration/questionnaire.spec.js#L67). It is getting into an infinite loop, but I don't know why.

Fatal RangeError: Maximum call stack size exceeded
    at trim_prefix (/home/vanessa/codes/Pantry-for-Good/node_modules/express/lib/router/index.js:288:23)
    at /home/vanessa/codes/Pantry-for-Good/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/home/vanessa/codes/Pantry-for-Good/node_modules/express/lib/router/index.js:335:12)
    at next (/home/vanessa/codes/Pantry-for-Good/node_modules/express/lib/router/index.js:275:10)
    at Layer.handle_error (/home/vanessa/codes/Pantry-for-Good/node_modules/express/lib/router/layer.js:67:12)
[...]

PR #286, opened 2017-10-27
creates donations - integration test

davycheung:
Per Issue #130 I've started some an integration test for donations. If this looks good, I'll continue with the rest of the donation tests.

### DIFF
--- a/server/controllers/donation.js
+++ b/server/controllers/donation.js
@@ -12,6 +12,11 @@ export default {
       total: req.body.items.reduce((acc, item) => acc + Number(item.value), 0)
     }
 
+    if (!req.user.roles.find(r => r === ADMIN_ROLE) &&
+        newDonation.donor !== req.user._id) {
+      throw new UnauthorizedError
+    }
+
     const donation = await Donation.create(newDonation)
     const donor = await Donor.findByIdAndUpdate(donation.donor,
       {$push: {donations: donation}},
@@ -48,8 +53,7 @@ export default {
 
   hasAuthorization(req, res, next) {
     if (req.user.roles.find(r => r === ADMIN_ROLE) ||
-        req.params.donationId === req.user._id ||
-        req.body.donor === req.user._id) {
+        req.params.donationId === req.user._id) {
       return next()
     }
 
--- a/server/controllers/donation.js
+++ b/server/controllers/donation.js
@@ -12,11 +12,6 @@ export default {
       total: req.body.items.reduce((acc, item) => acc + Number(item.value), 0)
     }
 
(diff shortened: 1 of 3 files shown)


kenjiO:
Looks great so far.  Please go ahead with the rest of them.

davycheung:
Cool, I'll continue working on them.

davycheung:
Updated with some additional donation tests.

server/controllers/donation.js
```diff
@@ -53,7 +48,8 @@ export default {
 
   hasAuthorization(req, res, next) {
     if (req.user.roles.find(r => r === ADMIN_ROLE) ||
-        req.params.donationId === req.user._id) {
+        req.params.donationId === req.user._id ||
+        req.body.donor === req.user._id) {
```
kenjiO: This can cause an issue since hasAuthorization is also used in sendEmail.  A malicious user can login with any user id.  Then he can send put requests to `/donations/:donationId` with body `{ donor: (his userId)}`.  This will cause hasAuthorization to pass and will allow him to access this endpoint for any `:donationId`.  Is there another way you can do this that doesn't open up this hole?

kenjiO: This can cause an issue since hasAuthorization is also used in sendEmail. A malicious user can login with any user id. Then he can send put requests to `/donations/:donationId` with body `{ donor: (his userId)}`. This will cause hasAuthorization to pass and will allow him to access this endpoint for any `:donationId`. Is there another way you can do this that doesn't open up this hole?

davycheung: Good point will look into.

davycheung: How do you feel about having a separate authorization function for creating donations? I.e. `canCreateDonation`

kenjiO: That sounds good to me.   When you make it I think it will be better to call the function from the `create(req, res)` controller method rather than putting it as a middleware in the routes file.

jspaine: There's already an auth check in the donation controller create method, admin or `donation.donor === user.id`. I think that's ok - admins can create donations for anyone, regular users can only create donations for themselves.

This `hasAuthorization` one looks weird though, `donationId === user.id`!? I think it should look up the donation from the db and check the donor is the same as the user. And just move it into the sendEmail function since that's the only place it's used.

davycheung: Made following commit to resolve the above issues:

https://github.com/freeCodeCamp/pantry-for-good/pull/286/commits/ca1ea1665e9e52f62b8fd1f66baeaab37670c42b

kenjiO:
Looks good now.  Thanks for the pull request

Issue #272, opened 2017-10-23
When changing customer status to inactive it sends the rejection email


kenjiO:
When an admin changes the status of a customer to inactive, the server sends the rejection email to the customer.  It should only send the rejection email when the status is changed to rejected.

prabhu-08:
@kenjiO I would like to work on this. Please assign it to me.

kenjiO:
OK, it's all yours prabhu

prabhu-08:
@kenjiO Raised a PR #288  for this. Let me know if this is fine.

PR #288, opened 2017-10-28
send rejected mail only if status is rejected

prabhu-08:
send rejected mail only if status is rejected Issue: #272


jspaine:
Looks good, thanks!

PR #290, opened 2017-10-28
Update README.md

makemommaproud:
Fixed grammar.


jspaine:
I prefer the original, sorry. I don't think ok vs okay matters and the extra comma isn't needed there.

Issue #300, opened 2017-10-31
Error on Packing List page when trying to pack with no preferred foods


kenjiO:
On the Packing List page, if there is a customer with no foods listed in the **Preferred Foods** column and you check the checkbox to the left of the # column, then click the green **Mark selected customers as packed** button, I get an error: **Package validation failed: contents: Path `contents` is required.**.   If there are no foods in the preferred foods column the checkbox in the first column of the table should be disabled.

jspaine:
This is done in #306, but the disabled checkbox state doesn't work well with the custom styled checkbox.

kenjiO:
I'm still getting this problem
![screen shot 2017-11-22 at 3 39 48 pm](https://user-images.githubusercontent.com/5035181/33154210-b09be77c-cf9b-11e7-8228-23a390790f21.png)

jcolborn-dropdeadgames:
I was reviewing this issue and I believe that this is only being seen in the seed data.  When I updated a customer and remove all of their Preferred Foods, I find that the checkbox is disabled.

kristoferma:
I looked at this issue and it has been solved in the current version.

![screen shot 2018-04-30 at 9 09 37 am](https://user-images.githubusercontent.com/9106939/39429746-dd7078e0-4c59-11e8-90c9-27bd54084aad.png)

The checkbox is disabled if the preferred foods section is empty. There seems to be some kind of mixup with the CSS input/checkbox styles.

I will look into it and try to make the disabled checkbox look better

PR #306, opened 2017-11-02
fix packingList

hy299792458:
Fix PackingList with unselectabler.

### DIFF
--- a/client/modules/food/components/packing/PackingList.js
+++ b/client/modules/food/components/packing/PackingList.js
@@ -107,7 +107,10 @@ class PackingList extends Component {
 
   handleSelectAll = isSelected => {
     const {selected} = this.state
-    const {customers} = this.props
+    const customers = []
+    this.props.packedCustomers.forEach((c) => {
+      if(c.foodPreferences.length > 0) customers.push(c)})
+
     this.setState({
       selected: isSelected || selected.length < customers.length ?
         customers.map(c => c._id) : []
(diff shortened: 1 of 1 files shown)


hy299792458:
Issue #300

client/modules/food/components/packing/PackingList.js
```diff
@@ -107,7 +107,10 @@ class PackingList extends Component {
 
   handleSelectAll = isSelected => {
     const {selected} = this.state
-    const {customers} = this.props
+    const customers = []
+    this.props.packedCustomers.forEach((c) => {
```
jspaine: This can be done a bit nicer with `Array.filter`, ie. `this.props.packedCustomers.filter(customer => ...`, but it's not really important. If you keep it like this though I'd declare customers with `let customers = []` to show you intend to change it. (The actual array the customers variable points to doesn't change if you push to it, so const works fine. I think it's a bit more descriptive though, and it's that way in the rest of the code. Same for `unselectable` below.)

client/modules/food/components/packing/PackingList.js
```diff
@@ -174,6 +179,7 @@ class PackingList extends Component {
               onSelect: this.handleSelect,
               onSelectAll: this.handleSelectAll,
               selected,
+              unselectable: unselectable,
```
jspaine: Can be just written as `unselectable` and expands to `unselectable: unselectable`, like `selected` in the line above.

hy299792458: Thanks, I’ll fix them.

PR #308, opened 2017-11-04
Fix/assign customers ui - fixes #294

### LINKED ISSUE #294: Assign customers UI improvement
On the Route Assignment page, I think it is confusing how it says Customers and then has the dropdown box with the driver name.  

screen shot 2017-10-30 at 1 36 18 pm  

In this example it looks to me like Neha Mayert is the customer.  I think we should change "Customers" to "Customers assigned to driver" or do some other change to the UI to make it more intuitive that what you are doing is using the select box to select a driver and listing the customers assigned to the driver selected.

logangingerich:
I made some minor adjustments to make the assign customers screen a bit more clear, but I'm open to suggestions on improving it further. 
screen shot 2017-11-04 at 3 10 24 pm
Changes:
* Uses Bootstrap's 'danger-alert' to highlight 'no matching customers'
* Adds a 'Customers' header to the list of assigned customers
* Changes 'Customers:' by the dropdown to 'Customers Assigned To:'


kenjiO:
It looks a lot more clear to me now. Thanks for contributing!

PR #312, opened 2017-11-09
Add missing settings prop to <RouteSummary>

### LINKED ISSUE #297: Suggest Route button on Route Assignment page not working
When I click this button it doesn't do anything and some elements of the page become unresponsive.  The browser console indicates "TypeError: Cannot read property 'getHostNode' of null"

wacii:
Suggest route was not working. This was because the preferred unit of distance was not being passed to the `<RouteSummary>` component via the `settings` prop.

`PropTypes` were defined on `<RouteSummary>` to prevent further problems.

As a side note, this problem was made more difficult to find because the error was being swallowed by [this line](https://github.com/freeCodeCamp/pantry-for-good/blob/9d54d5ebc95d0ba2f7e14251745024920cb93709/client/modules/driver/reducers/route.js#L100). Might want to consider re-throwing unexpected errors in these situations.

Closes #297

### DIFF
--- a/client/modules/driver/components/driver-assignment/RouteSummary.js
+++ b/client/modules/driver/components/driver-assignment/RouteSummary.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 
 const RouteSummary = ({route, settings}) => {
   const {distance, duration} = route.routes[0].legs.reduce((acc, x) => ({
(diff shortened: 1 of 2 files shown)


kenjiO:
Thanks for the fix wacii.

@jspaine , do you want to change the `dispatch(routeFailure(err)))` line that swallowed the error?

wacii:
Since you seemed interested, I went ahead and addressed the swallowed errors in PR #316.

PR #315, opened 2017-11-10
null check page.body before passing to react-quill

### LINKED ISSUE #292: Bug switching between page/email editors
The email and page editors share a lot of functionality so are implemented as one component with a prop to determine which should be rendered. It's quite broken when switching between the two.

It would be nice to keep it as one component but I'm not sure what the problem is. An easier fix would maybe be to separate it into two components.

thmsdnnr:
## the issue
When switching between Page and E-mail editors, we grab the current pages and stash them in state in `PageEditor`. This component's return method passes `page.body` to the ReactQuill component as the value to display in the textarea. However, if `page.body` is undefined, the ReactQuill component doesn't mount properly, giving an error like this:

screen shot 2017-11-10 at 8 18 06 am

When you then try to switch between the editor types, the body contents do not display properly, since the ReactQuill component holding the body text did not mount properly. The component receives all the updates (if you console.log out the page.body, the expected body shows up), but since ReactQuill failed to render once, it's "broken" and will not respond to any additional state changes. 

## caused by

The issue is that PageEditor, a child of EditPages, can have a condition where `this.state.page` exists but `this.state.page.body` is undefined. This seems to be due to a delay between fetching the props in the parent and then the way that we send the props to this component from its parent. In [L48-59 of PageEditor.js](https://github.com/freeCodeCamp/pantry-for-good/blob/staging/client/modules/page/components/EditPages.js#L48):

```javascript
  componentWillReceiveProps(nextProps) {
    if (this.props.loading && !nextProps.loading && !nextProps.error)
      this.setState({selectedPage: nextProps.pages[0].identifier})

    if (this.props.saving && !nextProps.saving && !nextProps.error)
      this.setState({dirty: false})

    if (nextProps.type !== this.props.type) {
      nextProps.loadPages()
      this.setState({selectedPage: null})
    }
  }
```
`props.type` holds the page type, one of `email` or `page`. When the page type changes (by clicking on the sidebar link), nextProps.type differs from `this.props.type`, so the third condition happens. PageEditor gets a `selectedPage` set to null, so it updates its state, since its `componentWillReceiveProps` looks like this:
```javascript
  componentWillReceiveProps(nextProps) {
    const {page, selectedPage} = nextProps

    if (selectedPage !== this.props.selectedPage) {
      this.setState({page: {...page}})
    }
  }
```
However, this new page has a page.body of `undefined`, presumably since `loadPages()` is still happening. If you console log a "1" "2" and "3" in each of the 3 conditions of PageEditor's `componentWillReceiveProps`, you get "1" the first time the page loads, and in between switching tabs on a given page, and then when you switch in between pages, you get a "3" and then a "1".

It's in this state between "3" and "1" where `page` exists and yet `page.body` is undefined and ReactQuill fails to mount.

## my solution
I experimented with a few approaches that alter the way `componentWillReceiveProps` works in the parent and child component. This got messy and I was worried I'd break something else in the process. Also, I felt like it'd be best not to couple the components too much in this way, since if we break them out into separate components down the road, it'd make that even messier.

I found the most effective solution to be simply to check for `page.body` being defined in the `return` of the `PageEditor`, and only mounting ReactQuill when we have a `value` to pass.

closes #292

### DIFF
--- a/client/modules/page/components/PageEditor.js
+++ b/client/modules/page/components/PageEditor.js
@@ -98,7 +98,7 @@ class PageEditor extends React.Component {
 
     return (
       <div className="page-editor">
-        {page &&
+        {page && page.body !== undefined &&
           <div>
             {type === pageTypes.EMAIL &&
               <div>


client/modules/page/components/PageEditor.js
```diff
@@ -115,15 +115,19 @@ class PageEditor extends React.Component {
...
-            <QuillBodyToolbar placeholders={bodyPlaceholders} type={type} />
-            <ReactQuill
-              theme="snow"
-              bounds=".page-editor"
-              value={page.body}
-              onChange={this.handleEditorChange(editorTypes.BODY)}
-              modules={getQuillBodyModules(type)}
-              ref={this.getEditorInstance(editorTypes.BODY)}
-            />
+            {page.body &&
```
kenjiO: I had no content on my Donor page so `page.body` was an empty string `""` which caused the editor not to show.  If I changed `page.body` to `page.body !== undefined` this fixed it.

thmsdnnr: Ah, thaks kenjiO, good catch! 

Does this solution seem kind of "hacky" to you as far as a fix? That's my main concern. I can look more into the way the component receives props and see if there's a way to make that happen in a way that page.body doesn't ever end up being undefined or "" so we wouldn't have to check at all, if you think that'd be better.

thmsdnnr: Looking into this more: it seems like the problem might be more in the parent EditPages component's `componentWillReceiveProps`. 

screen shot 2017-11-11 at 7 56 28 am

It seems like we're getting `selectedPage` out of sync with `pages` when we grab the new pages in EditPages. We ask PageEditor to grab a `selectedPage` with an identifier that doesn't yet exist in the state's `page` array.

In EditPages's `componentWillReceiveProps`, is there a way to make nextProps.loadPage() only call `setState({selectedPage})` after the new pages are loaded?

Does this seem like the right approach? We could avoid the intermediate state where selectedPage is null and ensure that we're only passing PageEditor a selectedPage with an identifier that we know for sure will exist.

jspaine: Great job on this, and nice explanations! That works so I think we should add it, can just do it on line 101 though - `{page && page !== undefined &&` and not need the extra div.

I think you're right that it's maybe not the best way and it would be nice to figure out the root cause. I'm pretty stumped, but that error's coming from the toolbar of the body editor and if you don't try to change its id (just call it 'body-toolbar' say in PageEditor's getQuillBodyModules and QuillBodyToolbar) it goes away, but the placeholders and buttons don't update.

> is there a way to make nextProps.loadPage() only call setState({selectedPage}) after the new pages are loaded?

You can just remove `this.setState({selectedPage: null})` on the line below but it doesn't help.

thmsdnnr: Cool — that's what I tried. I spent a really long time too spinning my wheels on trying to figure out how to refactor the componentWillReceiveProps and mapStateToProps of the parent and child component. One option for future reference might be making the PageEditor more of a "dumb component". Rather than selecting the selectedPage from the redux store itself, it could just receive whatever page the parent component has as currently selected. Then it could use local state to manage whatever editing tasks might be going on if necessary.

I added the line 101 check you suggested to the PR, although I did `page.body !== undefined`. If I just checked `page !==undefined`, I still got the `cannot set innerHTML of null` message.

Sorry for the extra commit: I forgot to take out the extra `div` & the page.body check lower down the page. Should be good now. Thanks for the patience :)

_Should I squash my commits and force-push these changes to tidy up? Would that be doing [this](https://stackoverflow.com/a/39334777)?:_
```
git checkout fix292
git rebase -i fix292~3 fix292

# at the interactive screen
# choose fixup for commit: 2 / 3

git push -u origin +fix292
```

thmsdnnr:
Thanks @kenjiO and @jspaine.

I added the line 101 check you suggested to the PR, although I did `page.body !== undefined`. If I just checked `page !==undefined`, I still got the `cannot set innerHTML of null` message.

Sorry for the extra commit: I forgot to take out the extra `div` & the `page.body` check lower down the page. I squashed my three commits and force-pushed to this branch. This is the first time I've done this, so please make sure I didn't mess it up or if it looks weird & I should fix something.

Thank you for the patience & your help 👍

kenjiO:
Everything looks OK to me with how you squashed and pushed it.  I'll go ahead and merge it in.

jspaine:
You could pass the page as a prop from the parent, it doesn't seem to help though. It should be the same anyway as passing it as a prop from react-redux.

Adding a default 'page' eg `{body: "a", subject: "a"}`, so state.page doesn't get set to undefined also doesn't help.

Adding another toolbar to the body editor, setting one as the 'page' toolbar with a static id and the other as the email toolbar, and only showing the appropriate one works fine.

You don't have to squash, we can do it while merging.

Thanks for looking into this!

thmsdnnr:
@jspaine No problem! Do you think I should still investigate this as a separate issue and try to do some sort of refactor? Or are you okay with the merge as-is?

jspaine:
It would be nice to know what the problem really is, so if you're interested go ahead. I don't know if it's a great use of time though unless you learn a lot from it.

PR #316, opened 2017-11-10
Log unexpected errors from `requestGoogleRoute()`

wacii:
`requestGoogleRoute()` currently swallows all errors. Better define expected errors rejected from `getGoogleRoute()` so all others can be rethrown.

Also removed an unnecessary try catch block from `getGoogleRoute()`.

### DIFF
--- a/client/modules/driver/reducers/helpers.js
+++ b/client/modules/driver/reducers/helpers.js
@@ -2,34 +2,35 @@ import {first, last} from 'lodash'
 import polyline from '@mapbox/polyline'
 
 const osrmUrl = 'https://router.project-osrm.org'
+const GOOGLE_ERROR_MAP = {
+  MAX_WAYPOINTS_EXCEEDED: 'Cannot find route for this many stops.',
+  OVER_QUERY_LIMIT: 'You have made too many requests for the current time period.'
+}
+const DEFAULT_GOOGLE_ERROR = 'There was an error calculating the route.'
 
 export const getGoogleRoute = (waypoints, optimize) => {
   const formattedWaypoints = waypoints.map(point => ({
     location: {lat: point.lat, lng: point.lng}
   }))
 
-  let resolve, reject
-  const result = new Promise((res, rej) => {
-    resolve = res
-    reject = rej
+  return new Promise((resolve, reject) => {
+    const directionsService = new google.maps.DirectionsService
+    directionsService.route({
+      origin: first(formattedWaypoints),
+      destination: last(formattedWaypoints),
+      waypoints: formattedWaypoints.slice(1, -1),
+      optimizeWaypoints: optimize,
+      travelMode: google.maps.TravelMode.DRIVING
+    }, (result, status) => {
+      if (status === google.maps.DirectionsStatus.OK) {
+        resolve(result)
+      } else {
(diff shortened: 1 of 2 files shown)


kenjiO:
@jspaine Does this look OK to you?

wacii:
Minor refactor added to last commit. I realized that without the try catch block, there was no reason to pull `resolve()` and `reject()` out of the Promise's context.

client/modules/driver/reducers/route.js
```diff
@@ -97,7 +97,12 @@ export const requestGoogleRoute = (waypoints, optimize) => dispatch => {
 
       dispatch(setWaypoints(optimizedWaypoints))
     })
-    .catch(err => dispatch(routeFailure(err)))
+    .catch(err => {
+      if (err.isDirectionsError)
+        dispatch(routeFailure(err.message))
+      else
+        throw err
```
jspaine: I think we should still `dispatch(routeFailure('There was a problem getting your route'))` or something here so it's visible to the user that something went wrong.

client/modules/driver/reducers/helpers.js
```diff
@@ -27,13 +20,12 @@ export const getGoogleRoute = (waypoints, optimize) => {
       if (status === google.maps.DirectionsStatus.OK) {
         resolve(result)
       } else {
-        reject(result)
+        const error = new Error(status)
```
jspaine: The status should be mapped to a string suitable to show to the user. I think `MAX_WAYPOINTS_EXCEEDED` should show an explanatory message. The rest could maybe just be a default message.

client/modules/driver/reducers/helpers.js
```diff
@@ -8,15 +8,8 @@ export const getGoogleRoute = (waypoints, optimize) => {
...
-  let resolve, reject
-  const result = new Promise((res, rej) => {
-    resolve = res
-    reject = rej
-  })
-
-  let directionsService
-  try {
-    directionsService = new google.maps.DirectionsService
+  return new Promise((resolve, reject) => {
```
jspaine, line 16: Yeah that's much cleaner. Could also do:
```js
return Promise.resolve().then(() => {
  if (ok) {
    return result  // instead of resolve(result)
  } else {
    throw error  // instead of reject(error)
  }
})
```

client/modules/driver/reducers/route.js
wacii: The caught error is likely programmatic, and unless an invalid value was passed to `google.maps.DirectionsService.route()`, it has nothing to do with getting the route. It could be a component missing a prop as in PR #312.

But if you don't want to have the app fail in such situations, I can change this PR to just fix the try catch block and improve how direction errors are reported.

jspaine: There's also the `MAX_WAYPOINTS_EXCEEDED` and maybe `OVER_QUERY_LIMIT` errors that could be useful to show. If it's a problem in the app I still think it's nicer to have a 'you clicked this and we tried but something went wrong' indication rather than a button that seems to do nothing.

I don't think it needs the try catch in `getGoogleRoute` though, just make an error with `isDirectionsError` like it is now but with an appropriate message, then in the catch of the `requestGoogleRoute` action creator dispatch a `routeFailure` with a generic error message if it's not a `isDirectionsError`.

jspaine:
Looks good, thanks!

PR #317, opened 2017-11-10
route reducer fix and show route errors

kenjiO:


client/modules/driver/reducers/route.js
```diff
@@ -206,7 +207,7 @@ export const createSelectors = path => {
 
   return {
     isFetching: state => get(state, path).routeFetching,
-    hasError: state => get(state, path).routeError,
+    hasError: state => get(state, path).error,
```
jspaine: The line above should just be `fetching` too shouldn't it?

kenjiO: Yes, thanks for spotting that one James.

PR #318, opened 2017-11-11
Improve food related authorization

### LINKED ISSUE #303: Schedule and Inventory access restrictions
If you create a volunteer account you can view data by going to these URLs
http://localhost:8080/schedule
http://localhost:8080/inventory

This data should only be accessible after an administrator has assigned the roles/inventory or roles/schedule to a volunteer.

wacii:
The INVENTORY and SCHEDULE roles were not being properly enforced, allowing any volunteer to perform related tasks.

* Restrict access to /inventory and /schedule based on role.

* Add more granular permissions to the food controller.

* Explicitly require authenticated user for all food endpoints.

I am not all that familiar yet with the code base, so the permissions I set up are likely not correct. But I did try to follow @jspaine's initial suggestions.

Closes #303

### DIFF
--- a/client/Router.js
+++ b/client/Router.js
@@ -2,7 +2,7 @@ import React from 'react'
 import {ConnectedRouter} from 'react-router-redux'
 import {Route} from 'react-router-dom'
 
-import {ADMIN_ROLE, clientRoles} from '../common/constants'
+import {ADMIN_ROLE, volunteerRoles} from '../common/constants'
 
 import Sidebar from './modules/core/components/sidebar/Sidebar'
 import Header from './modules/core/components/Header'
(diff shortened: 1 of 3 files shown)


client/Router.js
```diff
@@ -26,6 +26,8 @@ import SwitchWithNotFound from './components/router/SwitchWithNotFound'
 
 const IsAdmin = requireRole([ADMIN_ROLE])
 const IsVolunteer = requireRole([ADMIN_ROLE, clientRoles.VOLUNTEER])
+const canInventory = requireRole([ADMIN_ROLE, volunteerRoles.INVENTORY])
+const canSchedule = requireRole([ADMIN_ROLE, volunteerRoles.SCHEDULE])
```
jspaine: There's a `volunteerRoles.PACKING` too

server/controllers/food.js
```diff
@@ -101,9 +114,17 @@ export default {
...
    */
   async updateItem(req, res) {
+    authorizeByRole(req.user.roles, [INVENTORY, SCHEDULE])
+
     const originalCategoryId = req.params.foodId
     const updatedItem = req.body
 
+    const involvesCategoryChange = !updatedItem.categoryId || updatedItem.categoryId === originalCategoryId
+    if (involvesCategoryChange) {
+      const cantInventory = !intersection(req.user.roles, [ADMIN_ROLE, INVENTORY]).length
```
jspaine: Can't it just be `if (involvesCategoryChange) authorizeByRole(req.user.roles, [INVENTORY])`?

client/Router.js
wacii: Added a check for that as well. `IsVolunteer()` is no longer used, only more specific checks.

server/controllers/food.js
wacii: Yes it can. Changed.

client/Router.js
jspaine: Thanks!

PR #319, opened 2017-11-12
Fix/converts dates to short format - fixes #299

### LINKED ISSUE #299: Format the dates on the Packing List page
On the Packing List page the **Last Packed On** and **Date Packed** columns show dates like this **2017-10-30T18:37:14.553Z**.  It would be nice to have them in an easier to read format like 10/30/2017

logangingerich:
I used Moment.JS to convert dates on the packing list to a more readable, short format (MM/DD/YYY). 
However, if anyone else is still working on this, please feel free to disregard and drop. I just wanted to take a crack at it and see if I could get it working.


client/modules/food/components/packing/PackingList.js
```diff
@@ -259,6 +260,15 @@ function exists(thing) {
   return thing
 }
 
+function isValid(date) {
```
jspaine: Maybe call this `formatDate` instead. Also the else below should be on the same line as both the braces - `} else {`

logangingerich: Thanks, I’ll make the changes as soon as possible! I’ll try to address the commit issues as wel, not sure what happened there.

logangingerich: Just made the suggested changes. Sorry about the weird commits, I'm not sure how they got screwed up but, as you suggested, I'll delete and re-create my fork after this is merged. Hopefully that will give me a blank slate. Thanks!

PR #321, opened 2017-11-15
Add istanbul test coverage

kenjiO:
With these changes `npm run coverage` will generate a test coverage report in `coverage/lcov-report/index.html`

The one thing I don't like is that I think the babel-plugin-istanbul plugin also gets used with `npm test` when you want to run just the tests without creating a coverage report.  From what I understand, istanbul adds extra code to every statement to keep track of what statements run.  I don't like the idea of all that unnecessary code being put into regular test runs.  

If you have any ideas on how to resolve this or if this shouldn't be a concern please let me know.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,7 +9,7 @@
   ],
   "env": {
     "test": {
-      "plugins": ["babel-plugin-rewire"]
+      "plugins": ["istanbul", "babel-plugin-rewire"]
     }
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ node_modules/
 typings/
 server/config/env/secrets.js
 dist
+.nyc_output/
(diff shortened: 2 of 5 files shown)


jspaine:
Sweet! What is the coverage!?

How about just `NODE_ENV=test_coverage` and move it to a separate environment in babelrc?

kenjiO:
The coverage is around 40%

I tried using `NODE_ENV=test_coverage` but then some of the tests failed.  I think some of the tests depend on NODE_ENV being set to test.

jspaine:
Yeah I think so, I'll just merge it for now then.

kenjiO:
OK, if you notice the tests are running slower or any other unusual behavior lets just roll this back and apply it temporarily when we want to check the coverage

PR #324, opened 2017-11-19
fixes #243 - validates that donation is NaN on server

### LINKED ISSUE #243: Donation Values Treated as Strings
Looks like the donation values are being concatenated instead of being summed up.

screen shot 2017-09-21 at 9 59 03 pm
screen shot 2017-09-21 at 9 59 34 pm

logangingerich:
Checks that donation value is NaN on the server, and throws a BadRequestError if so.


jspaine:
Thanks!

PR #325, opened 2017-11-20
Logout on expired sessions

### LINKED ISSUE #296: Client not indicating session timeout
The server will timeout an inactive session after an hour, but if the user does not refresh the page after the timeout, it will not indicate that the session has expired and will have errors and missing data for anything fetched from the api server.   After the timeout, It will be good if the client can go to the login page, or perform some other behavior so users realize their login session is expired.

Note: to shorten the timeout on the server for testing set the timeout in milliseconds on [this line](https://github.com/freeCodeCamp/pantry-for-good/blob/de1bc1777b4040690e9e999a74f11986e47b8b03/server/config/env/all.js#L14).   
  
Example of what happens when trying to see the packing page after the session expires:
screen shot 2017-10-30 at 2 24 41 pm

wacii:
If the user continues using the app after having timed out, all requests will fail, some silently. This PR instead clears the client session when the server signals a session timeout.

The main change was simple. In the api middleware, if a response has status 401, clear the user from client state. b1cbc57

To support this change the server needed to return 401 responses consistently when the user was not authenticated. 7262004

On logging out an admin, the formatting was off; there was an empty space where the sidebar used to be. This is because of DOM manipulation being done in `<Application />` using _jQuery_ instead of _React_. c0c9105.

The last change wasn't strictly necessary, it was a simple refactor I did while reviewing how the client handled authorization. I thought it might be useful in better enforcing guest only pages, but I didn't find anywhere else that I was certain needed it. Other than maybe `<ResetPassword />`. 3b48bc3

Closes #296.

### DIFF
--- a/client/components/router/guestOrRedirect.js
+++ b/client/components/router/guestOrRedirect.js
@@ -1,0 +1,42 @@
+import React from 'react'
+import {connect} from 'react-redux'
+import {push} from 'react-router-redux'
+
+import selectors from '../../store/selectors'
+import userClientRole from '../../lib/user-client-role'
+
+const mapStateToProps = state => ({
+  user: selectors.auth.getUser(state)
+})
+
+const mapDispatchToProps = dispatch => ({
+  push: location => dispatch(push(location))
+})
+
+const redirectIfAlreadySignedIn = props => {
+  if (props.user && props.user._id) {
+    const role = userClientRole(props.user)
+    props.push(role ? `/${role.split('/')[1]}s` : '/')
+  }
+}
+
+const guestOrRedirect = WrappedComponent => {
+  class GuestOrRedirect extends React.Component {
+    constructor(props) {
+      super(props)
+      redirectIfAlreadySignedIn(props)
+    }
+  
+    componentWillReceiveProps(nextProps) {
+      redirectIfAlreadySignedIn(nextProps)
+    }
+  
+    render() {
+      return <WrappedComponent {...this.props} />
+    }
+  }
+
+  return connect(mapStateToProps, mapDispatchToProps)(GuestOrRedirect)
+}
+
+export default guestOrRedirect
--- a/server/routes/donation.js
+++ b/server/routes/donation.js
@@ -1,5 +1,8 @@
 import Router from 'express-promise-router'
 import donationController from '../controllers/donation'
+import usersController from '../controllers/users'
+
+const {requiresLogin} = usersController 
 
 const donationRouter = Router({mergeParams: true})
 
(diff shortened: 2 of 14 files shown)


wacii:
A question. Some of the express routes were exported as factory methods. So...

```js
import buildDonorRouter from "server/routes/donor";
const donorRouter = buildDonorRouter();

// as opposed to just...

import packingRouter from "server/routes/packing";
```

It seemed connected to whether or not the router was pulling methods off of `userRouter`, but I couldn't figure out why this was done.

client/components/router/guestOrRedirect.js
```diff
@@ -0,0 +1,43 @@
...
+
+const guestOrRedirect = WrappedComponent => {
+  class GuestOrRedirect extends React.Component {
+    constructor(props) {
+      super(props)
+      redirectIfAlreadySignedIn(props)
+    }
+  
+    componentWillReceiveProps(nextProps) {
+      if (this.props.user !== nextProps.user)
```
jspaine: Won't this be true if the user updates their profile too?

jspaine:
Thanks, it looks good!

The route factories were for testing, didn't get round to doing more than the [api router](https://github.com/freeCodeCamp/pantry-for-good/blob/staging/server/tests/unit/api.spec.js) though.

client/components/router/guestOrRedirect.js
wacii: It was an attempt to limit the number of times `redirectIfAlreadySignedIn()` was called, but now that I think about it, Redux `connect()` is already doing this. I'll remove it.

It would be true, but I don't see that being a problem. If the component hadn't redirected a user away from the sign in page at that point, the `redirectIfAlreadySignedIn()` should be called again.

wacii:
I removed `if (this.props.user !== nextProps.user)`.

server/routes/donation.js
```diff
@@ -1,5 +1,8 @@
 import Router from 'express-promise-router'
 import donationController from '../controllers/donation'
+import usersController from '../controllers/users'
```
kenjiO, line 3: Can this be simplified to `import { requiresLogin } from 'usersController'`.    And when it occurs in packing.js, questionnaire.js, settings.js as well.

wacii: Hmm, I don't think so. The curly braces in imports are not exactly destructuring statements, but a way to pull named exports from the module. With the default export being a special case, but still just a named export.

wacii: Although I have never tried it, so I suppose it could fall back to destructuring the default export.

client/components/router/guestOrRedirect.js
jspaine: Oh yeah it's only used for pages that shouldn't be accessible by logged in users, ignore me!

PR #326, opened 2017-11-22
fixes #323 - aligns selectors in navbar in Firefox

### LINKED ISSUE #323: UI Problem- Navbar dropdown icon is not at same height with navbar item
screen shot 2017-11-17 at 9 29 13 pm

Operating System: Mac high Sierra 11.0
Browser: Mozilla Firefox

logangingerich:
Removes float: right and instead uses absolute positioning, which seems to fix the Firefox-specific UI problem.

### DIFF
--- a/client/modules/core/css/core.css
+++ b/client/modules/core/css/core.css
@@ -29,7 +29,8 @@ li.treeview > a > span::after {
   font-size: 10px;
   font-weight: 300;
   padding-right: 10px;
-  float: right;
+	position: absolute;
+  right: 0px;
   content: '\f078';
 }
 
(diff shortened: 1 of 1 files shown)


client/modules/core/css/core.css
```diff
@@ -29,7 +29,8 @@ li.treeview > a > span::after {
   font-size: 10px;
   font-weight: 300;
   padding-right: 10px;
-  float: right;
+	position: absolute;
```
kenjiO: Can you change the tab indentation to 2 spaces?  Thanks!

logangingerich: No idea why it indented like that, but I’ll fix it up!

PR #327, opened 2017-11-22
Prevent deleting customers with packages

### LINKED ISSUE #275: Packages component breaks if a customer is deleted
If a package is packed for a customer, then the customer is deleted, when you go to the packages page it will not load the list of Packed Packages and the console indicates **TypeError: Cannot read property '_id' of undefined**.  The error happens in [this line](https://github.com/freeCodeCamp/pantry-for-good/blob/297cf2035ad7c6d8d4498af34848c04d2082aa9f/client/modules/food/components/packing/Packages.js#L68).  Although it can be fixed simply by checking if customer is undefined I think it will be better to also prevent deleting a customer if any packages have a reference to that customer.  Instead of deleting it just set a flag indicating the customer is deleted but keep it in the database.  Most likely a food bank will want to keep a record of the customers for all the packages they send out.

kenjiO:
closes #275

### DIFF
--- a/server/controllers/customer.js
+++ b/server/controllers/customer.js
@@ -3,6 +3,8 @@ import {extend, intersection, includes} from 'lodash'
 import {ForbiddenError, NotFoundError} from '../lib/errors'
 import {ADMIN_ROLE, clientRoles, volunteerRoles, customerStatus} from '../../common/constants'
 import Customer from '../models/customer'
+import Volunteer from '../models/volunteer'
+import Package from '../models/package'
 import mailer from '../lib/mail/mail-helpers'
 import User from '../models/user'
 import {updateFields} from '../lib/update-linked-fields'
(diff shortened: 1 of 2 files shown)


server/controllers/customer.js
```diff
@@ -70,10 +71,17 @@ export default {
   async delete(req, res) {
     const id = req.customer._id
 
-    await User.findByIdAndRemove(id)
-    await Customer.findByIdAndRemove(id)
+    // Check to see if the customer is associated with any packages
+    const packages = await Package.find({customer: id})
```
jspaine: There's also a `count` method that just returns the number of matches directly.

jspaine:
Looks good, thought you were going for the `deleted` flag though? But I guess that would only be useful if you wanted to delete the customer and leave the package?

kenjiO:
Here it is using count()
I also added a statement to delete the customer from any volunteers it is assigned to. 

In the future I still would like to have either a deleted or retired flag for the customer to indicate the customer is no longer with the food bank but to keep its data for record keeping.  But for now this prevents packages from referencing a customer that no longer exists.

PR #332, opened 2017-11-29
Capitalized the column heading status on the Packing List page

vanessasena:
On the Packing List page the column heading 'status' was changed to 'Status' so it is capitalized like the other table headings.
This was suggested by @kenjiO


kenjiO:
This looks perfect.  Thanks!

PR #338, opened 2017-12-01
Add admin toggle to edit user page

wacii:
This adds the ability to promote/demote users to/from admin status. Admins can't demote themselves. Although they're not prevented from trying, only told after the fact that they can't with an error message.

Could use some input on design/where exactly you want this feature. I'm real bad at design.

### DIFF
--- a/server/controllers/users/profile.js
+++ b/server/controllers/users/profile.js
@@ -1,5 +1,6 @@
-import {extend} from 'lodash'
+import {extend, includes} from 'lodash'
 
+import {ADMIN_ROLE} from '../../../common/constants'
 import {BadRequestError} from '../../lib/errors'
 import User from '../../models/user'
 
@@ -33,7 +34,7 @@ export const getById = async function(req, res) {
 export const update = async function(req, res) {
   let user = req.user
   if (user._id !== req.body._id)
-    user = await User.findById(req.body._id).lean()
+    user = await User.findById(req.body._id)
 
   // For security measurement we remove the roles from the req.body object
   delete req.body.roles
@@ -49,6 +50,19 @@ export const update = async function(req, res) {
   const sameEmail = await User.findOne({email: req.user.email}).lean()
   if (sameEmail && sameEmail._id !== req.user._id)
     throw new BadRequestError('Email address is taken')
+
+  // Update admin status
+  if (includes(req.user, ADMIN_ROLE)) {
+    const alreadyAdmin = includes(user.roles, ADMIN_ROLE)
+    if (user.isAdmin && !alreadyAdmin) {
+      user.roles.push(ADMIN_ROLE)
+    } else if (!user.isAdmin && alreadyAdmin) {
+      if (parseInt(req.params.userId, 10) === req.user._id)
+        throw new BadRequestError('You are not allowed to demote yourself')
+      user.roles.splice(user.roles.indexOf(ADMIN_ROLE), 1)
+    }
+    delete user.isAdmin
+  }
 
   await user.save()
   res.json(user)
--- a/client/modules/users/components/EditUser.js
+++ b/client/modules/users/components/EditUser.js
@@ -1,7 +1,9 @@
 import React from 'react'
 import { connect } from 'react-redux'
 import { Col, Row } from 'react-bootstrap'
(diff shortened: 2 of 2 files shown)


server/controllers/users/profile.js
```diff
@@ -61,3 +62,32 @@ export const me = async function(req, res) {
...
+  user.roles.push(ADMIN_ROLE)
+  await user.save()
+  return res.json(user)
+}
+
+/**
+ * Take Admin priveleges away from User
+ */
+export const demote = async function(req, res) {
+  if (req.params.userId === req.user._id)
```
jspaine: `req.params.userId` is a string but `req.user._id` is a number

jspaine:
I'm not great at design either, but maybe it could be a checkbox in with the rest of the form?

server/controllers/users/profile.js
wacii: Oh right. I forgot about that. Fixed.

wacii:
The toggle is now a checkbox. Remember thinking earlier that it would be better to have the toggle separate from the edit user form, but I can't remember why, so it must not have been a very good reason.

This PR is now dependent on #341. Something I had to fix to get this to work.

jspaine:
Cool, it looks a bit weird with the label above and nothing next to the checkbox though, you can add text next to the checkbox with `options={"the text"}`.

It also complains `user.save is not a function` because in user profile update, user can just be an object not a mongoose model.

wacii:
Oh right. `.lean()` returns a plain javascript object. My mistake.

I removed the label and put text next to the checkbox.

client/modules/users/userReducer.js
```diff
@@ -34,6 +34,24 @@ export const editUser = user => ({
...
+    endpoint: `admin/users/${userId}/demote`,
+    method: 'PUT',
+    schema: userSchema,
+    types: [actions.SAVE_REQUEST, actions.SAVE_SUCCESS, actions.SAVE_FAILURE]
+  }
+})
+
+export const promoteUser = userId => ({
+  [CALL_API]: {
+    endpoint: `admin/users/${userId}/promote`,
```
kenjiO: I don't think this endpoint exists yet in the server routes

wacii:
I finished cleaning up the old routes.

client/modules/users/userReducer.js
wacii: It was there before, but when I merged the behavior into `profile.update()` I didn't fully cleanup the old routes.

server/controllers/users/profile.js
```diff
@@ -50,6 +51,17 @@ export const update = async function(req, res) {
   if (sameEmail && sameEmail._id !== req.user._id)
     throw new BadRequestError('Email address is taken')
 
+  // Update admin status
```
kenjiO, line 54: Lines 54-62 should only be ran if `user` has the `ADMIN_ROLE`.

wacii: Right. I'll need to update `<EditUser>` as well. Forgot all of this was being used for two different purposes.

kenjiO: There is already a check for admin in the `UserRouter.js` file that loads `<EditUser>`.  Currently this component is only being used for admins to edit user accounts.  So I think it would be OK to just do the check on the server side for now.

server/controllers/users/profile.js
```diff
@@ -50,6 +51,19 @@ export const update = async function(req, res) {
...
+  if (includes(req.user, ADMIN_ROLE)) {
+    const alreadyAdmin = includes(user.roles, ADMIN_ROLE)
+    if (user.isAdmin && !alreadyAdmin) {
```
jspaine, line 57: I think this is pretty confusing. `req.user` is the original user, `user` is the original user updated with `req.body`, `isAdmin` is a prop set on the new user... I think it would be easier to follow if the admin role was added/removed on the client and the controller just compared hasRole(oldUser, admin) with hasRole(newUser, admin).

Sorry for the delay, I'll merge it like this because it does make sense, but I looked at it a couple of days ago and couldn't really follow it in the time i had so held off on it. It's really not bad though, just think it could be a bit clearer.

PR #340, opened 2017-12-03
Updated installing API Keys section on README

vanessasena:
Updated Installing API Keys section, because there is a script executed after installation that copies the secrets.js and it is not necessary to copy the file manually


kenjiO:
Thanks for catching that.  It looks like the instructions didn't get updated when that step was automated.

PR #342, opened 2017-12-05
Fix server error when updating a user

kenjiO:
the call to lean() is removing the user.save function which is required later in the function.  Also added some tests since this was not being caught by them.


kenjiO:
I just saw that #338 fixes this.  Please still merge this for the tests.  If there is a merge conflict I can delete the changes to the controller.

Issue #157, opened 2017-07-04
add a 'you signed up with your google account' email


jspaine:
and in the users/password controller just return 200 so the client can say  'an email has been sent to {whatever address they entered}'

mtrivera:
@jspaine Can I work on this one?

ronaldblanco:
Hello @jspaine 
Hey let me help you with this one, give me a little more of background of what you need.

kenjiO:
When a user signs up through google [this function](https://github.com/freeCodeCamp/pantry-for-good/blob/06a0322e0b3261ef30720e82dcb0c9e36d8cd259/server/config/strategies/google.js#L35-L56) gets called.  You can add something here that sends an email saying the person signed up using their google account.  Methods for sending an email are in [server/lib/mail/mail-helpers.js](https://github.com/freeCodeCamp/pantry-for-good/blob/staging/server/lib/mail/mail-helpers.js).  You can use the send method or create another method similar to the other methods in this file.

ronaldblanco:
Hey @kenjiO 
Went I try to Sing Up to test I have this error:
![image](https://user-images.githubusercontent.com/7390703/33674703-3c230e6e-da7e-11e7-8155-8ed5fc0bec65.png)
And it is not related to any change I did.

vanessasena:
I also had this problem before. Try going to:
https://console.developers.google.com/apis/library/plus.googleapis.com
and see if the API is enabled for you. If not, click to enable it.



Att,

--
*Vanessa Sena Teixeira*

ronaldblanco:
Thanks!

ronaldblanco:
Hello to everyone;
I did a pull request to close this issue, when you think everything it is ready I will then squash into only one commit.

Let me know.

PR #343, opened 2017-12-07
(#157) add a 'you signed up with your google account' email 

ronaldblanco:
Closes the Issue #157 add a 'you signed up with your google account' email.
All changes were locally tested and lint checked.

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -54,7 +54,8 @@ export const pageIdentifiers = {
   CUSTOMER_UPDATED: 'customer-update',
   DONATION_RECEIVED: 'donation-received',
   DONATION_RECEIPT: 'donation-receipt',
-  PASSWORD_RESET: 'password-reset'
+  PASSWORD_RESET: 'password-reset',
+  PASSWORD_RESET_GOOGLE: 'password-reset-google'
 }
 
 export const pageTypes = {
--- a/server/config/mailer.js
+++ b/server/config/mailer.js
@@ -49,5 +49,6 @@ export default async function sendEmail(
     path: '/v3/mail/send',
     body: mail.toJSON()
   })
+  
   return sg.API(request)
 }
--- a/server/controllers/users/password.js
+++ b/server/controllers/users/password.js
@@ -5,8 +5,8 @@ import config from '../../config'
 import {
   BadRequestError,
   NotFoundError,
-  UnauthorizedError,
-  ValidationError
+  UnauthorizedError
+  // ValidationError
 } from '../../lib/errors'
 import mailer from '../../lib/mail/mail-helpers'
 import User from '../../models/user'
@@ -26,23 +26,19 @@ export const forgot = async function(req, res) {
   // Lookup user by email
   if (!req.body.email) throw new BadRequestError('Email is required')
   const user = await User.findOne({email: req.body.email}).select('-salt -password')
-  if (!user) {
-    throw new ValidationError({email: 'No account with that email has been found'})
-  } else if (user.provider !== 'local') {
-    return res.status(400).send({
-      message: 'It seems like you signed up using your ' + user.provider + ' account'
-    })
-  } else {
+  if (user && user.provider !== 'local') {
+    if (user.provider === 'google') await mailer.sendPasswordGoogle(user)
+  } else if (user && user.provider === 'local') {
     user.resetPasswordToken = token
     user.resetPasswordExpires = Date.now() + 3600000 // 1 hour
 
+    const port = process.env.NODE_ENV === 'production' ? '' : `:${config.port}`
+    const url = `${config.protocol}://${config.host}${port}/users/reset-password/${token}`
+
+    await mailer.sendPasswordReset(user, url)
     await user.save()
   }
 
-  const port = process.env.NODE_ENV === 'production' ? '' : `:${config.port}`
-  const url = `${config.protocol}://${config.host}${port}/users/reset-password/${token}`
-
-  await mailer.sendPasswordReset(user, url)
   res.send({message: 'Password reset email sent'})
 }
 
--- a/server/lib/seed/seed-data.js
+++ b/server/lib/seed/seed-data.js
@@ -185,6 +185,17 @@ export const pages = [{
 <p>Thanks for your donations to:</p>
 <p><span class="ql-placeholder-content" data-id="receipt" data-label="Donation Receipt" data-required="true"></span></p>
 `
+}, {
+  identifier: pageIdentifiers.PASSWORD_RESET_GOOGLE,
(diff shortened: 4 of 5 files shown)


jspaine:
Hi, it looks good! This one was a bit complicated though because I'm not sure the email needs to be editable, it should just say something like 'you appear to have signed up with google, try logging in with google. if you have any problems send us a message at support@foodbank.email' and can just have default translations (when we get to that). Or what do you think?

Another part of this issue was that the server users password controller should just send a 200 regardless, and the client should show 'an email has been sent to email@address', so the route can't be used to figure out registered users email addresses. You don't have to do this too, I was a bit late to the discussion on the issue but thought it was worth mentioning.

ronaldblanco:
Hello @jspaine; I do not understand what do you mean with: "can just have default translations (when we get to that). Or what do you think?" I did search in the other emails and I do not see any translations options implemented. It is something new do you want me to code?

jspaine:
Hi, I just meant that this email is so simple it's not really worth being editable, and there could just be a standard one. But it makes the code a bit more complicated to make it not editable. I guess we shouldn't worry about it for now and leave it editable.

I'll leave some suggestions in the code, but the context of this is a user enters their email in the forgot password form but they signed up with google so don't have a password and the password reset email wouldn't be appropriate. So I don't think the place you're sending the email from is right, look for how the forgot password form is handled on the server.

common/constants.js
```diff
@@ -54,7 +54,8 @@ export const pageIdentifiers = {
   CUSTOMER_UPDATED: 'customer-update',
   DONATION_RECEIVED: 'donation-received',
   DONATION_RECEIPT: 'donation-receipt',
-  PASSWORD_RESET: 'password-reset'
+  PASSWORD_RESET: 'password-reset',
+  GOOGLE_SINGUP: 'google-singup'
```
jspaine: should be signup

server/lib/seed/seed-data.js
```diff
@@ -185,6 +185,14 @@ export const pages = [{
 <p>Thanks for your donations to:</p>
 <p><span class="ql-placeholder-content" data-id="receipt" data-label="Donation Receipt" data-required="true"></span></p>
 `
+}, {
+  identifier: pageIdentifiers.GOOGLE_SINGUP,
+  title: 'Google Sined Up',
+  type: pageTypes.EMAIL,
+  subject: '<p>The User Sined Up with Google!</p>',
```
jspaine: This goes to users who think they forgot their password but can actually just sign in with google. They don't want to be called `The User`! I'd put something like:
`<Foodbank Name>` password reset request
Hi `<User Name>`,
A request was made to reset your password for `<Foodbank Name>` but your account is linked to your google account. You can just login by clicking the 'Sign in with Google' button.
If you still cannot login you can email our support team at `<Foodbank Support Email>`
If you didn't make this request you can ignore this email

common/constants.js
ronaldblanco: It is not better a more specific name? tomorrow you could have another sing-up too.

ronaldblanco:
Hey @jspaine take a look at this.

common/constants.js
jspaine: I mean it's spelled signup. A better name though might be password-reset-google, because it's more related to password reset than signup.

ronaldblanco: I Get it, the problem was that the first code was done in another context.
I will fix it.

ronaldblanco:
Check now.

server/config/mailer.js
```diff
@@ -49,5 +49,7 @@ export default async function sendEmail(
     path: '/v3/mail/send',
     body: mail.toJSON()
   })
+  // console.log(toEmail + '->' + content)
+  // console.log(content)
```
jspaine: Remember to remove

server/controllers/users/password.js
```diff
@@ -29,6 +29,7 @@ export const forgot = async function(req, res) {
   if (!user) {
     throw new ValidationError({email: 'No account with that email has been found'})
   } else if (user.provider !== 'local') {
+    if (user.provider === 'google') await mailer.sendPasswordGoogle(user)
```
jspaine: Exactly, but now it can just send this mail if user.provider is google, or save the reset token and send the regular password reset email if user.provider is local. 

Then just `res.send({message: 'Password reset email sent'})` for all cases (though it would be nice to show an error if there's a problem actually sending the email)

ronaldblanco: Sorry I do not get it, do you want me to do something "to show an error if there's a problem actually sending the email"?
The rest it is working as you are describing.

jspaine: It's not though -
- if there's no matching user it throws `ValidationError({email: 'No account with that email has been found'})` 
- if `user.provider = google` it sends `{message: 'It seems like you signed up using your google account'}`
- otherwise it sends `{message: 'Password reset email sent'}`

It should always send `{message: 'Password reset email sent'}` unless something goes wrong.

ronaldblanco:
Check now, if it is ok I will clean the code and squash.

server/config/mailer.js
ronaldblanco: OK

server/controllers/users/password.js
```diff
@@ -29,9 +29,10 @@ export const forgot = async function(req, res) {
   if (!user) {
     throw new ValidationError({email: 'No account with that email has been found'})
```
jspaine: Almost, it shouldn't throw an error if there's no user found though

server/controllers/users/password.js
```diff
@@ -29,9 +29,10 @@ export const forgot = async function(req, res) {
...
   } else if (user.provider !== 'local') {
-    return res.status(400).send({
+    if (user.provider === 'google') await mailer.sendPasswordGoogle(user)
+    /*return res.status(400).send({
       message: 'It seems like you signed up using your ' + user.provider + ' account'
-    })
+    })*/
   } else {
     user.resetPasswordToken = token
     user.resetPasswordExpires = Date.now() + 3600000 // 1 hour
```
jspaine: and the mailer.sendPasswordReset has to be in this else block, or it'll be called for google users too

ronaldblanco:
Check now!

jspaine:
Try to think through it more, the if...else's can be made a bit simpler now. We only want to send an email if there's a user.

mailer.sendPasswordReset needs the password reset url too.

ronaldblanco:
And now??

jspaine:
Yeah that looks good.

ronaldblanco:
It should be ready!
Let me know.

jspaine:
Thanks, looks good!

Issue #337, opened 2017-11-30
possible issue with extend(user, req.body)


kenjiO:
https://github.com/freeCodeCamp/pantry-for-good/blob/686a83b1c3374b465d33c21d151a954163340ceb/server/controllers/users/profile.js#L30-L40

Is the `extend(user, req.body)` statement here a problem if someone changes the _id attribute sent with req.body or some other attribute of req.body? Except for the roles property, it looks like the user can send whatever they want as req.body and it will overwrite the req.user.

wacii:
I don't know MongoDB well enough to say what all could happen, but regardless I would suggest whitelisting the keys you wanted to have updated. So something like this:

```js
extend(user, pick(req.body, ['email', 'firstName', 'lastName']))
```

jspaine:
Yeah this is definitely a problem. Could also use omit if a model has lots of fields but only a few need to be restricted. Mongoose won't save any fields not defined in the schema.

It's going to be a bit annoying to keep things in sync if fields need to be added. Also some controllers might need to allow different fields depending on if the user is admin/other roles and could get a bit messy. Maybe the routes would be a better place to do it?

PR #344, opened 2017-12-07
Filter params in controller

wacii:
Instead of dumping whatever the user gives you into a model, selectively take or reject specific keys.

I was a little uncertain about the client controllers. I wrote something for the create method, but I imagine that `email`, `firstName`, and `lastName` could be omitted as well, instead pulling those values off of the related user. Unless there is a reason to allow those values to diverge.

I didn't touch the update method. It's probably fine as the authorizations are restricted to admins and the clients themselves, but it's worth some thought.

Addresses #337, not sure if it resolves it though.

### DIFF
--- a/server/controllers/customer.js
+++ b/server/controllers/customer.js
@@ -1,4 +1,4 @@
-import {extend, intersection, includes} from 'lodash'
+import {extend, intersection, includes, omit} from 'lodash'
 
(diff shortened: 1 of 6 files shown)


jspaine:
Thanks! The duplication of firstName, email... is a remnant from the old app, I think it makes sense to not have to fetch the user for every client that's loaded but we probably do need to think about how these are kept in sync.

PR #345, opened 2017-12-11
Fix profile controller update method

wacii:
The recently added param filtering removed `isAdmin` from user, which prevents the admin toggle from working.

Also fixed the cannot demote self check. Was using `userId` from params, but against convention, the id is not included in the path but the request body.

The admin toggle code was refactored for clarity and a test was added.

Now that I look at it, I had to change the authenticated user is admin check as well. Not sure how this was working before...

### DIFF
--- a/server/controllers/users/profile.js
+++ b/server/controllers/users/profile.js
@@ -49,16 +49,20 @@ export const update = async function(req, res) {
     throw new BadRequestError('Email address is taken')
 
   // Update admin status
-  if (includes(req.user, ADMIN_ROLE)) {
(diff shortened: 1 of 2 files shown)


jspaine:
Huh, yeah that's a bit weird the put route has no userId param. Didn't you add the `parseInt(req.params.userId` because it was a string and `req.user._id` was a number though?!?

wacii:
I guess I was just assuming? The code ended up getting changed a number of times, so I suppose I just lost track of what had and had not been tested. I really should have just added some tests.

PR #346, opened 2017-12-12
Limit route requests

wacii:
These are time-consuming and expensive requests, so I limited them to one outgoing.

I hard-coded the selector in [the route "reducer"](https://github.com/freeCodeCamp/pantry-for-good/compare/staging...wacii:feature/limit-route-requests?expand=1#diff-c891cd2901f7fe7ee3be201fef407506R88). The route selector factory is defined in the same file, so I can't reference the ultimate selectors without pulling apart the reducer file.

### DIFF
--- a/client/modules/driver/reducers/route.js
+++ b/client/modules/driver/reducers/route.js
@@ -83,7 +83,11 @@ export const moveWaypoint = (waypoint, idx, getWaypoints) => (dispatch, getState
   // )
 }
 
-export const requestGoogleRoute = (waypoints, optimize) => dispatch => {
+export const requestGoogleRoute = (waypoints, optimize) => (dispatch, getState) => {
+  // TODO: this should be using selectors, but the needed ones are defined here
+  if (getState().delivery.route.fetching)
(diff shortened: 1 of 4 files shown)


client/modules/driver/reducers/route.js
```diff
@@ -83,7 +83,11 @@ export const moveWaypoint = (waypoint, idx, getWaypoints) => (dispatch, getState
   // )
 }
 
-export const requestGoogleRoute = (waypoints, optimize) => dispatch => {
+export const requestGoogleRoute = (waypoints, optimize) => (dispatch, getState) => {
+  // TODO: this should be using selectors, but the needed ones are defined here
+  if (getState().delivery.route.fetching)
+    return Promise.reject(new TypeError('Already requesting a route'))
```
jspaine: It's not a type error, why not just Error?

wacii: @jspaine Changed it.

jspaine: Thanks!

Issue #347, opened 2017-12-13
Non-admin users able to update other users


kenjiO:
The following method is allowing non-admin users to update details for other users.  Only users with `ADMIN_ROLE` should be able to update other users.  If a user without `ADMIN_ROLE` role sends a request to this endpoint with another user's id in `req.body._id` then it should return a 403 Forbidden error.

https://github.com/freeCodeCamp/pantry-for-good/blob/43bd74183cae89c7ed12a83fa57f5516f814e693/server/controllers/users/profile.js#L31-L70

wacii:
Yeah, this should really be two separate methods: a user updating their profile, and an admin updating another user. It's tempting to merge them, because they share so much logic, but then you inevitably end up in such situations.

I'll handle this one.

kenjiO:
There is a test for this [here](https://github.com/freeCodeCamp/pantry-for-good/blob/025a5599435a549be32db0ec5fec54970d1169be/server/tests/integration/user.spec.js#L475) that I set to pending until this is fixed.  When you fix this can you change `xit` to `it` so the test is no longer pending.  Thanks.

wacii:
@kenjiO I noticed that. I moved the endpoint under '/admin', so only admins can access it. So I don't think the test is necessary anymore.

PR #348, opened 2017-12-13
Server tests for updating users

kenjiO:


server/tests/integration/user.spec.js
```diff
@@ -412,5 +412,173 @@ describe('User Api', function() {
...
+
+      await userReq.put('/api/users').send(requestBody).expect(200)
+
+      const updatedUser = await User.findById(user._id).lean()
+      expect(updatedUser).to.have.property('email', requestBody.email)
+      expect(updatedUser).to.have.property('firstName', requestBody.firstName)
+      expect(updatedUser).to.have.property('lastName', requestBody.lastName)
+    })
+
+    xit('regular users cannot update another user profile', async function(){
```
kenjiO, line 475: This test will fail until #347 is fixed so I set it as pending using `xit`.  After #347 is fixed then it can be included.

PR #349, opened 2017-12-15
Revise customer integration tests

wacii:
Separated the tests into endpoints to better see what was and wasn't being tested.

Removed a duplicate read customer route and moved the list customer route under admin. That is how it was being used anyways.

### DIFF
--- a/client/modules/customer/reducer.js
+++ b/client/modules/customer/reducer.js
@@ -11,7 +11,7 @@ export const actions = crudActions('customer')
 
 export const loadCustomers = () => ({
   [CALL_API]: {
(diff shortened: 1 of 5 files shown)


wacii:
Not sure why this fails. The failing tests are nowhere near the max timeout on my machine. I'll look into this.

jspaine:
Looks good. Not sure you can do much other than raise the timeout limit, they don't even seem to be new tests.

wacii:
Closing/reopening to trigger Travis build.

I didn't notice, but you're right. The failing tests aren't even new.

PR #350, opened 2017-12-15
Fix promoting user bug

### LINKED ISSUE #347: Non-admin users able to update other users

wacii:
A normal user with knowledge of the application could promote others to become admins.

This was addressed by pulling apart the controller function, which was previously shared by two endpoints; one for updating your own profile, and another for admins to update other users.

While this introduces some duplication, it is easier to understand, with less room for these sorts of bugs.

Closes #347


wacii:
I let admins update anything on a user but `_id`, because that would be weird.

client/modules/users/authReducer.js
```diff
@@ -32,7 +32,7 @@ export const clearFlags = () => ({type: CLEAR_FLAGS})
 
 export const setProfile = user => ({
   [CALL_API]: {
-    endpoint: 'users',
+    endpoint: 'users/me',
```
jspaine, line 35: I don't know about reusing the me endpoint for this, seems a bit inconsistent with resource/:resourceId for updates too?

server/tests/integration/user.spec.js
```diff
@@ -464,81 +463,14 @@ describe('User Api', function() {
...
-      await userReq.put('/api/users').send(requestBody).expect(200)
+      await userReq.put('/api/users/me').send(requestBody).expect(200)
 
       const updatedUser = await User.findById(user._id).lean()
       expect(updatedUser).to.have.property('email', requestBody.email)
       expect(updatedUser).to.have.property('firstName', requestBody.firstName)
       expect(updatedUser).to.have.property('lastName', requestBody.lastName)
     })
 
-    xit('regular users cannot update another user profile', async function(){
```
jspaine: I think some of these tests, and in the next file, could be useful to keep, maybe it's obvious from the code now that these issues can't happen but it would be good to have some assurance that they'll never happen.

client/modules/users/authReducer.js
wacii: They are supposed to represent two different resources. Your own profile, accessible to everyone, and an arbitrary user, accessible only to admins. Resources don't have to map one to one with database objects...or documents, or whatever they're called in mongo.

Now that I look at it, `userRouter.route('/users/:userId').get(users.getById)` should be admin only also. No reason to let normal users fetch other users' data, especially with sequential ids.

But if you don't like the idea, I could solve #350 otherwise.

server/tests/integration/user.spec.js
wacii: Sure, I'll leave in some as regression tests. Should I add a comment referencing the bug in the code? I would normally just go ahead with it, but the contributors guidelines here are a little restrictive in how issue references can be used.

jspaine: It's up to you. I'm not sure why the guidelines say not to reference the issue in the PR.

client/modules/users/authReducer.js
jspaine: Ah yeah sorry, and if it's only for updating your own profile it's pointless to pass the id.

Good catch on the user get.

Issue #122, opened 2017-06-15
Show notifications of users applied/updated, donations, delivery route changed


jasonsreynolds:

jspaine:
Currently it's set up to email the admin when a customer applies or updates their application. It would be much better with a notification system to handle these events.

Maybe users could have an array of notifications sub-documents and use a pre-save hook to push notifications to users by role?

ronaldblanco:
Hello @jspaine let me help you with this.
Give me and idea of what you are needing here.

Thanks.

jspaine:
Hi, something like the notifications on github or facebook. Some examples:
- if a user applies admins should get a notification like "user name applied as a customer"
- clicking the notification should take you to the customers application
- there should be a button to clear the notification
- if a customer changes their address or status the assigned driver should get a notification

It could get complicated but I think supporting those would be a good start.

ronaldblanco:
This notifications will be in the model, save in mongodb right??

jspaine:
Yeah

ronaldblanco:
Hola @jspaine I am having a issue at the staging branch I am not able to create a local user (for tests) it just do nothing and do not create the user.

jspaine:
Hi @ronaldblanco, how are you trying to create a user? You should be able to manually sign up at least to create users.

ronaldblanco:
I think I know what it is happening, the problem is: when the system does not create the user (possible validation problems) it does not rise any message or alert.
I did try now again benign sure to fulfill all possible validation and the user was create correctly.

ronaldblanco:
Hey @jspaine do you want me to make the pull request so you can go checking the progress of this issue?
Or, do I finish all first?

ronaldblanco:
Hey @jspaine  could you please check this gist an advice me?
https://gist.github.com/ronaldblanco/6a7a072823eef4a28ba4a7456cc093a1

<NotificationLinkItem path={notification.url} title="Notification URL" />
//Warning: Failed prop type: The prop `path` is marked as required in `NotificationLinkItem`, but its value is `undefined`.

"notification.url" it is a string for all the RFFieldGroup objects but it is constantly undefined for the link I am trying to create.
Probably this is a silly issue but I am stock on it and I do not found any difference.

Thanks.

jspaine:
What npm packages were you thinking of? 
[...]

PR #353, opened 2018-01-02
(#122) Show notifications of users applied/updated, donations, delive…

ronaldblanco:
…ry route changed
Closes or completed #122, section limited to customers creation and modification (as a start).
Notification structure create in mongodb for every user.
All changes tested locally and lint corrected.

### DIFF
--- a/client/modules/users/authReducer.js
+++ b/client/modules/users/authReducer.js
@@ -2,6 +2,12 @@ import {get} from 'lodash'
 
 import {CALL_API} from '../../store/middleware/api'
 
+import {denormalize} from 'normalizr'
+import {createSelector} from 'reselect'
+import { arrayOfNotifications, notification as notificationSchema } from '../../../common/schemas'
+import { crudActions} from '../../store/utils'
+export const actions = crudActions('auth')
+
 export const CLEAR_USER = 'auth/CLEAR_USER'
 export const CLEAR_FLAGS = 'auth/CLEAR_FLAGS'
 export const SET_PROFILE_REQUEST = 'auth/SET_PROFILE_REQUEST'
@@ -25,10 +31,39 @@ export const SIGNIN_FAILURE = 'auth/SIGNIN_FAILURE'
 export const LOAD_REQUEST = 'auth/LOAD_REQUEST'
 export const LOAD_SUCCESS = 'auth/LOAD_SUCCESS'
 export const LOAD_FAILURE = 'auth/LOAD_FAILURE'
+export const LOAD_NOTIFICATION_REQUEST = 'auth/LOAD_NOTIFICATION_REQUEST'
+export const LOAD_NOTIFICATION_SUCCESS = 'auth/LOAD_NOTIFICATION_SUCCESS'
+export const LOAD_NOTIFICATION_FAILURE = 'auth/LOAD_NOTIFICATION_FAILURE'
 
 export const clearUser = () => ({type: CLEAR_USER})
 
 export const clearFlags = () => ({type: CLEAR_FLAGS})
+
+export const loadNotifications = () => ({
+  [CALL_API]: {
+    endpoint: '/users/me/notifications',
+    schema: arrayOfNotifications,
+    types: [LOAD_NOTIFICATION_REQUEST, LOAD_NOTIFICATION_SUCCESS, LOAD_NOTIFICATION_FAILURE]
+  }
+})
+
+export const deleteNotification = id => ({
+  [CALL_API]: {
+    endpoint: `/users/me/notifications?id=${id}`,
+    method: 'DELETE',
+    schema: notificationSchema,
+    types: [actions.DELETE_REQUEST, actions.DELETE_SUCCESS, actions.DELETE_FAILURE]
+  }
+})
+
+export const deleteAllNotifications = () => ({
+  [CALL_API]: {
+    endpoint: `/users/me/notifications?id=`,
+    method: 'DELETE',
+    schema: notificationSchema,
+    types: [actions.DELETE_REQUEST, actions.DELETE_SUCCESS, actions.DELETE_FAILURE]
+  }
+})
 
 export const setProfile = user => ({
   [CALL_API]: {
@@ -91,6 +126,31 @@ export const loadUser = () => ({
   }
 })
 
+const removeNotification = (state, actionIndex) => {
+  const notifications = [
+    ...state.user.notifications.slice(0, actionIndex),
+    ...state.user.notifications.slice(actionIndex + 1),
+  ]
+  return {
+    ...state,
+    user: {
+      ...state.user,
+      notifications
+    }
+  }
+}
+
+const removeAllNotification = state => {
+  const notifications = []
+  return {
+    ...state,
+    user: {
+      ...state.user,
+      notifications
+    }
+  }
+}
+
 export default (state = {user: null}, action) => {
   switch (action.type) {
     case CLEAR_USER:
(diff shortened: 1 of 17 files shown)


jspaine:
Good start! I'd try and remove the redux form stuff though, you shouldn't need it.

You also shouldn't get notifications for things you did, only things other users did, and 'viewing' a notification should delete it.

I think the notification page and dropdown could be combined, I guess I'd go with the dropdown and add a 'remove all' button at the bottom and remove buttons on the side of each notification. I think AdminLTE or font awesome has an icon you can use instead of the word notifications.

Or if you keep the notifications page instead I think the remove button(s) should just delete the notification(s) without having to click save

ronaldblanco:
I will be checking your suggestions, but concerning "You also shouldn't get notifications for things you did, only things other users did" do the application have a way to control which user did a modification? Because I do no see that capability right now.

jspaine:
On the server you can just look at `req.user` to get the user making the request, then just add notifications to all other (relevant) users.

ronaldblanco:
Hey @jspaine could you help me here.
how can I use a props function for save the notification in a:
onClick={ () => {fields.remove(index)} } of the FieldArray?
I want something like onClick={ () => {fields.remove(index)} && save } do the remove and then save; I had tried a lot of options but I am stock.

Thanks.

jspaine:
Hi, why are you using a FieldArray though? It's for rows of inputs where you need to store the value of each input, validation states etc. The notifications don't have any inputs, they can just be `<li>`'s with an onClick to go to the url and a button to delete. 

The api should just have GET to return all the users notifications and DELETE to delete one or an array of notifications. There's no need for the client to ask to create notifications since the server should know when to do it.

If you do those everything will be much easier and you can use some of the same patterns used in other places for the client state that'll make deleting easy.

Just storing notifications on users should be enough, because volunteers are users too.


Sorry it's so much to do! It's a pretty big issue on both the front and back end. I'd probably focus on getting the server right or simplifying the display components first so it's a bit more manageable. Let me know if there's anything else.

ronaldblanco:
OK, I will be working on this and I let you know.
Thanks

ronaldblanco:
Hello @jspaine;

Hey I have been a little busy on my job so I am slow on this one.
I need your advise, I am trying to create the server controller but I am unable to test the results even with postman, when I try to check the api, example: GET "localhost:3000/users" I always get the error:
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8">
        <title>Error</title>
    </head>
    <body>
        <pre>Cannot GET /users</pre>
    </body>
</html>
Could you tell me how to work and test efficiently the server apis in this application.

Thanks.

jspaine:
Hi, no problem.

The api routes are mounted under the api prefix, see `server/config/express`:
`app.use('/api', apiRoutes(...))`

Though for many of them you'll also need to send a cookie to show you're authenticated. So the easiest way to test it is probably writing an integration test like the ones in `server/tests/integration`.

The steps are:
- in before, call initDb() to set up the db and clear any collections your tests are going to use
- in after and afterEach, clean up any collections used and reset the db

for each test use the `server/test/helpers` to:
- create a user and an app that bypasses the session (cookie) authentication check for that user
- set up supertest to use that app
- use supertest to call the route that runs your controller method
- check the route response is what you expect

ronaldblanco:
Thanks.

ronaldblanco:
Hello @jspaine 
Hey when you have a time take a look to the last commit; I have the server side ready and the only it is not working on the client side (file: NotificationsNew.js) it is the auto-refresing process when I delete a notification.
Could you advice me on this, I did used the DonorList.js component as a template.

Thanks.

ronaldblanco:
Hello @jspaine 
Take a look now!

Thanks.

ronaldblanco:
"Just storing notifications on users should be enough, because volunteers are users too."

I did set the notifications to volunteers because I notice that some of then were not in the users collection.

client/modules/users/authReducer.js
```diff
@@ -30,6 +36,32 @@ export const clearUser = () => ({type: CLEAR_USER})
 
 export const clearFlags = () => ({type: CLEAR_FLAGS})
 
+export const loadNotifications = () => ({
+  [CALL_API]: {
+    endpoint: '/users/me/listnotifications',
```
jspaine: Couldn't these all use the same endpoint, `users/me/notifications/:id?`? GET should return all notifications, DELETE could delete by id or delete all if no id is passed. I'm not sure why you'd need to update a notification so the setProfileNotifications function below could probably be removed

client/modules/users/authReducer.js
```diff
@@ -91,6 +132,18 @@ export const loadUser = () => ({
   }
 })
 
+const rmNotification = (state, actionIndex) => {
+  const newState = state
```
jspaine: This will mutate the state, the state var is a reference to the state, so in this line you assign the same reference to newState. So both state and newState reference the same data and changing one will change the other. So you have to do something like this:
```
const notifications = [
  ...state.user.notifications.slice(0, index),
  ...state.user.notifications.slice(index + 1),
]

return {
  ...state,
  user: {
    ...state.user,
    notifications
  }
}
```

client/modules/users/authReducer.js
```diff
@@ -91,6 +132,18 @@ export const loadUser = () => ({
...
+  newState.user.notifications.splice(actionIndex, 1)
+  return newState
+}
+
+const rmAllNotification = state => {
+  const newState = state
```
jspaine:
```
return {
  ...state,
  user: {
    ...state.user,
    notifications: []
  }
}
```

client/modules/users/authReducer.js
```diff
@@ -30,6 +36,32 @@ export const clearUser = () => ({type: CLEAR_USER})
 
 export const clearFlags = () => ({type: CLEAR_FLAGS})
 
+export const loadNotifications = () => ({
+  [CALL_API]: {
+    endpoint: '/users/me/listnotifications',
+    schema: arrayOfNotifications,
+    types: [actions.LOAD_ALL_REQUEST, actions.LOAD_ALL_SUCCESS, actions.LOAD_ALL_FAILURE]
```
jspaine: These are actions for load/delete a user and shouldn't be reused for this in case something else is listening for them, just add some new ones to the top of the file

ronaldblanco:
Hey @jspaine,  check now and let me know.

ronaldblanco:
Hey @jspaine, are you there?

jspaine:
Hi, yeah sorry, I tried it before and a notification wasn't generated and I forgot to get back to you. It's looking pretty good, there are some bits you could clean up like the commented out code, the searchVolunteerAndSetNotifications. I'm not really a fan of the abbreviations, LOAD_NOTI_, rmNotifications. The notifications list component file is still called notificationsNew.

Some tests would be useful, and if you want to do more on this it would be nice if you could define with the routes that calls to that endpoint should send notifications to users with x roles.

I need to look into why the tests are failing on travis.

ronaldblanco:
Hello @jspaine 
Hey I did some fix and cleaning according to your suggestions.
In the case of:
"Some tests would be useful, and if you want to do more on this it would be nice if you could define with the routes that calls to that endpoint should send notifications to users with x roles."
Could you give me a little more of detailed, I don't understand what do you mean with "calls to that endpoint should send notifications"; or maybe lead me to an example in the app.

jspaine:
Hi, sorry I forgot about this, you can send a message to remind me if I don't respond for a while.

I mean if you call PUT /api/customer/customerId it should send a notification that the customer was updated to admins. So if you set up a test with an admin user and a customer user, and update the customer, then you should be able to get the notifications for the admin and see the customer updated notification.

There's an example of a similar process in the customer integration test - `server/tests/integration/customer.spec.js`, line 211 -  `updates other customers if admin`, you'd probably need to create the admin user before updating the customer though.

jspaine:
Hi, no problem, been really busy too.

I was thinking the customer would update themselves, and the admin should get the notification. Then you only have to make 2 users.

ronaldblanco:
Hello @jspaine 
Can you give a look to the test I did for the customer creation and modification?
image
The problem it is that the first time I run the command "npm run test:server" always fails but them I commented it and begin to play with it and start working perfect, but always running the command the first time it fails.
Could you take a look and let me know?

Thanks.

ronaldblanco:
Hello @jspaine 

Hey take a look to the test I did on the customer file I think it is according to the expectations; the code needs to be clean so let me know to clean and rebase.

Thanks.

ronaldblanco:
Hello???

ronaldblanco:
Hello @jspaine 

Hey I did a cleaning and a re-base to make the revision more easy for you (I think)

jspaine:
Hi, thanks, it seems to be working great! I'll merge it now and open an issue to keep track of the remaining work.

ronaldblanco:
Excellent!
Thanks.

PR #354, opened 2018-01-03
Added column Customer name to Packages grid

### LINKED ISSUE #352: Add Customer Name to Packed Packages List
Would be useful to add the Customer name to the packed packages list. The customer column is currently only displaying the Customer ID. Good one for a first-timer. 

screen shot 2018-01-01 at 5 00 15 pm

sneh1:
closes #352


jspaine:
Thanks, looks good!

PR #355, opened 2018-01-04
Update FoodCategories.js

Bubblemelon:
Resolving the order of category items on the inventory page. 

Details: 
Category items are alphabetically ordered, newly added categories are also alphabetized (not placed at the bottom of the inventory list). Lowercases have higher precedence over uppercase letters.

### DIFF
--- a/client/modules/food/components/inventory/FoodCategories.js
+++ b/client/modules/food/components/inventory/FoodCategories.js
@@ -29,8 +29,26 @@ export class FoodCategories extends Component {
   }
 
   getTableRows = () => {
-    if (this.props.foods.length > 0) {
-      return this.props.foods.map(category =>
+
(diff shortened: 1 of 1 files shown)


jspaine:
Thanks, couple of points though:

`Array.sort` alters the array it's called on. You shouldn't change a react components props from inside the component or react will re-render it thinking the input (prop) has changed, which could cause the prop to be changed again etc.

Instead of writing the comparator function yourself, there's a `String.localeCompare` ([example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Sorting_non-ASCII_characters)).

`sorted` would be a better name than `sorting`, what's assigned to it is the sorted array.

PR #357, opened 2018-01-09
Refactor FoodItems into two components

kenjiO:
I think the FoodItem.js component is doing too much and the code is hard to reason with.  So I kept the functionality for listing the food items in that file and moved all the functionality dealing with the form for adding and editing foods into a new FoodAddEditForm component.


jspaine:
Yeah, that looks much better.

PR #358, opened 2018-01-27
Add s-alert notifications

### LINKED ISSUE #154: alerts to show errors/success
react-s-alert looks ok and is also used by react-bootstrap-table (in v4, and i'm not sure we need alerts from tables)

DavOnGit:
Notify on submitting form field with errors and on settings saved succesfully.
It uses a redux middleware that can be very flexible, all notifications logic is in one file.

This PR resolves #154

### DIFF
--- a/client/Application.js
+++ b/client/Application.js
@@ -1,6 +1,7 @@
 import React, {Component} from 'react'
 import {connect} from 'react-redux'
 import PropTypes from 'prop-types'
+import Notify from 'react-s-alert'
 
 import {ADMIN_ROLE} from '../common/constants'
 import selectors from './store/selectors'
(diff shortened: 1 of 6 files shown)


kenjiO:
This looks really nice.  There is one minor change I think will make it look better.  The way the alert displays now makes it not stand out from the rest of the form.
screen shot 2018-01-30 at 9 30 13 am


I tried it with `<Alert stack={{limit: 3}} position={'top'}  html={true} timeout={5000} effect={'flip'}/>` and it is a lot more noticable to me.  Can you do it this way instead?

screen shot 2018-01-30 at 9 51 53 am

DavOnGit:
All done @kenjiO 
You should see `...position={'top'}, offset={0}...` props in _Application.js_ file.

PR #359, opened 2018-02-01
Feature - Add Tooltip to Packed Package List Preferred Foods

### LINKED ISSUE #356: Packed package list create a 'contents' tool-tip
The "Contents" column on the packed packages list is not really useful. Their are usually a number of contents which are being cut off. Perhaps someone can scrap the text column and shift the contents into a modal that is activated by a button on the right most column next to "Unpack".

screen shot 2018-01-06 at 6 11 07 pm

caleblogan:
closes #356 

The module for the table `react-bootstrap-table` doesn't have a nice way of attaching listeners to cells. I wrapped the contents of each cell with an `OverlayTrigger`. Not sure if this adds too much extra markup.


kenjiO:
Looks good.  Thanks!

PR #361, opened 2018-02-05
Created Button to mark a Package As Received

mbetances805:
Creating this PR to review work in progress.

I created a button to mark a package from Packed to Received. I could not figure out how to enable just one column in the table to be editable with react-bootstrap-table. Including 'cellEdit' would make the entire table editable.

Question: Do the individual items in the package also need to be updated to 'Received? At this time only the package is updated to 'Received'.

I had an issue committing my code due to the 'admins can set other users as admins' test. That test kept failing, so I commented it out.

Let me know if this is in the right direction.

Thanks!


client/modules/food/components/packing/Packages.js
```diff
@@ -39,14 +53,37 @@ class Packages extends Component {
...
       <Button bsStyle="danger"
         onClick={() => this.props.unpack(foodPackage._id)}
+        disabled={foodPackage.status === 'Received'}
       ><i className="fa fa-undo" /> Unpack</Button>
-    </div>
-
+    )
+  }
+    
+  getPackageButton = (_, foodPackage) => {
```
jspaine: Why not do this in `getActionButtons` above? Then they can be in the same table column

client/modules/food/reducers/packing.js
```diff
@@ -142,6 +159,21 @@ export default (state = {}, action) => {
         saving: false,
         saveError: null,
       }
+    case RECEIVED_REQUEST: {
```
jspaine: You can add this under `case UNPACK_REQUEST` above, then all request type actions will set the loading state. Same with the failure action below.

client/modules/food/reducers/packing.js
```diff
@@ -142,6 +159,21 @@ export default (state = {}, action) => {
...
+      return {
+        ...state,
+      }
+    }
+    case RECEIVED_SUCCESS:
+      return {
+        ...state,
+        ids: without(state.ids, action.response.packages._id).concat(action.response.packages),
```
jspaine: The ids don't change, so all you'd need to do here is clear saving and error states

client/modules/food/components/packing/Packages.js
```diff
@@ -8,27 +8,41 @@ import {BootstrapTable, TableHeaderColumn, SizePerPageDropDown} from 'react-boot
...
 
 const mapDispatchToProps = dispatch => ({
   load: () => dispatch(listPackages()),
-  unpack: packageId => dispatch(unpackPackage(packageId))
+  unpack: packageId => dispatch(unpackPackage(packageId)),
+  received: singlePackage => dispatch(receivedPackage(singlePackage)),
 })
 
 class Packages extends Component {
+  constructor() {
```
jspaine: If you make the changes I commented in the reducer you shouldn't need to keep any state or reload the packages

server/controllers/packing.js
```diff
@@ -14,7 +14,14 @@ export default {
   list: function(req, res) {
     Package.find().then(data => res.json(data))
   },
-
+  complete: async function(req, res) {
+    const singlePackageId = req.body.singlePackage
```
jspaine: I'd just call it `packageId`, if it was more than a single one it would be an array

client/modules/food/components/packing/Packages.js
```diff
@@ -8,27 +8,41 @@ import {BootstrapTable, TableHeaderColumn, SizePerPageDropDown} from 'react-boot
 import 'react-bootstrap-table/dist/react-bootstrap-table.min.css'
 
 import selectors from '../../../../store/selectors'
-import {listPackages, unpackPackage} from '../../reducers/packing'
+import {listPackages, unpackPackage, receivedPackage} from '../../reducers/packing'
```
jspaine: As the food bank, the packages are delivered not received, I'd change the names in the code and ui text

mbetances805:
@jspaine I have made the changes per your feedback. Please review and let me know if any further changes are required. Thanks!

jspaine:
Code looks good! The buttons don't quite fit for me though and it still says 'Received' (just saw the status also says Received), want to fix it or shall I just merge this?

mbetances805:
@jspaine I was following what I saw in the model for status. I can update the model to have the status as 'Delivered' instead of 'Received' and update the code elsewhere to reflect that change. I'll also look into the responsiveness of the two buttons under one column.

mbetances805:
@jspaine, apologies for the delay. I updated the status and variable names from 'Received' to 'Delivered'. I addressed the responsive issue with the two buttons in one column. Please review and let me know if any further updates are required.

client/modules/food/components/packing/Packages.js
```diff
@@ -84,9 +109,7 @@ class Packages extends Component {
             <TableHeaderColumn dataField="status" >
               Status
             </TableHeaderColumn>
-            <TableHeaderColumn
-              dataFormat={this.getActionButtons}
-            />
+            <TableHeaderColumn dataFormat={this.getActionButtons} />
```
kenjiO: The unpack and delivered buttons are not fitting on my browser either.  Can you change this line to `<TableHeaderColumn dataFormat={this.getActionButtons} width="200px"/>`?  That should make the column big enough so they fit side by side.

client/modules/food/components/packing/Packages.js
```diff
@@ -30,6 +32,12 @@ class Packages extends Component {
     this.props.load()
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props.packSaving !== nextProps.packSaving) {
+      this.props.load()
```
jspaine: You don't need to reload if you send the data back in the right format, the api middleware will handle updating the store. See below

client/modules/food/reducers/packing.js
```diff
@@ -44,6 +48,19 @@ export const pack = newPackedCustomers => dispatch => {
     .catch(error => dispatch(packFailure(error)))
 }
 
+export const deliveredPackage = singlePackage => dispatch => {
```
jspaine: This can just be:
```js
export const deliveredPackage = packageId => ({
  [CALL_API]: {
    endpoint: 'packing',
    method: 'PUT',
    body: {packageId},
    responseSchema: packageSchema,
    types: [DELIVERED_REQUEST, DELIVERED_SUCCESS, DELIVERED_FAILURE]
  }
})
```

The api returns a package, it normalizes it with the package schema and updates the package entities in the store, so the ui updates.

server/controllers/packing.js
```diff
@@ -14,7 +14,14 @@ export default {
   list: function(req, res) {
     Package.find().then(data => res.json(data))
   },
-
+  complete: async function(req, res) {
+    const packageId = req.body.singlePackage
```
jspaine: `const {packageId} = req.body`

server/controllers/packing.js
```diff
@@ -14,7 +14,14 @@ export default {
...
+    const deliveredPackage = await Package.findByIdAndUpdate(packageId, {status: 'Delivered'}, {new: true})
+    if (!deliveredPackage) {
+      throw new BadRequestError(`package with _id ${req.body.singlePackage} not found`)
+    }
+    res.json({packages: deliveredPackage})
```
jspaine: `res.json(deliveredPackage)`

client/modules/food/reducers/packing.js
mbetances805: @jspaine I made the updates, but it seems that I need to change the responseSchema to arrayOfPackages for the UI to update. Please confirm if you agree and I'll wrap up the updates.

jspaine: Ah, it should work if you respond with single package from the controller, are you sure you changed the response in the controller to `res.json(deliveredPackage)` rather than `res.json({packages: deliveredPackage})`?

Otherwise just push it and I'll take a look. Thanks!

mbetances805:
@jspaine, made updates requested. please review.

jspaine:
Looks good! Thanks!

PR #365, opened 2018-03-13
Fix error 'unathorized' while an user wants to 'Apply as a customer'

### LINKED ISSUE #362: Register as a customer shows not authorized from trying to get foods
GET foods should be unrestricted

gomezvillegasdaniel:
The user interface tries to populate a table with the food list, but at that point the user does not have any permission yet (required to get the food list), that's why the error was being displayed.
closes #362

### DIFF
--- a/server/controllers/food.js
+++ b/server/controllers/food.js
@@ -1,4 +1,4 @@
-import {extend, intersection} from 'lodash'
+import {extend, intersection, some} from 'lodash'
 
 import {
   BadRequestError,
@@ -6,11 +6,10 @@ import {
   NotFoundError,
   ValidationError
 } from '../lib/errors'
-import {ADMIN_ROLE, clientRoles, volunteerRoles} from '../../common/constants'
+import {ADMIN_ROLE, volunteerRoles} from '../../common/constants'
 import Customer from '../models/customer'
 import Food from '../models/food'
 
-const {CUSTOMER} = clientRoles
 const {INVENTORY, SCHEDULE} = volunteerRoles
 
 export default {
@@ -70,7 +69,7 @@ export default {
    * List of Food categories
    */
   async list(req, res) {
-    authorizeByRole(req.user.roles, [CUSTOMER, INVENTORY, SCHEDULE])
+    authorizeByRole(req.user.roles)
 
     const foods = await Food.find()
       .sort('category')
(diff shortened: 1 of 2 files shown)


gomezvillegasdaniel:
@QuincyLarson Hi, I'd like to get a review, greetings

server/controllers/food.js
```diff
@@ -70,7 +69,7 @@ export default {
    * List of Food categories
    */
   async list(req, res) {
-    authorizeByRole(req.user.roles, [CUSTOMER, INVENTORY, SCHEDULE])
+    authorizeByRole(req.user.roles)
```
kenjiO, line 72: Thanks for your pull request.  Since the issue states that this route should be unrestricted the only change needed would be to remove this line.  However I think it would be good to at least restrict this route to logged on users.  If a user is logged on that means that `req.user` will be set.  So instead of calling `authorizeByRole` I would just check if `req.user` is set, and if not throw the unauthorized error.

server/controllers/food.js
```diff
@@ -1,16 +1,15 @@
-import {extend, intersection} from 'lodash'
+import {extend, intersection, some} from 'lodash'
 
 import {
   BadRequestError,
-  ForbiddenError,
+  UnauthorizedError,
```
gomezvillegasdaniel: @kenjiO I think this should be changed to UnathorizedError

server/controllers/food.js
gomezvillegasdaniel: @kenjiO  yes, it sounds better, thanks, hands on it

server/controllers/food.js
```diff
@@ -160,9 +159,14 @@ export default {
   }
 }
 
+function isAdmin(role) {
+  return role === ADMIN_ROLE
+}
+
 function authorizeByRole(userRoles, roles = []) {
-  if (!intersection(userRoles, [...roles, ADMIN_ROLE]).length) {
-    throw new ForbiddenError
```
gomezvillegasdaniel, line 165: @kenjiO I think this should be changed to UnathorizedError

server/controllers/food.js
jspaine: It has to be unrestricted because it's used to show the food preferences selector on the customer signup form

server/controllers/food.js
jspaine: If there is a user but they don't have the required roles it should be forbidden, if there's no user it should be unauthorized. The `requiresLogin` middleware used in the routers handles throwing unauthorized errors

gomezvillegasdaniel: ok, thanks

server/controllers/food.js
gomezvillegasdaniel: ok, thanks

gomezvillegasdaniel: @kenjiO I noticed that the route... is already covered by requiresLogin?
https://github.com/freeCodeCamp/pantry-for-good/blob/5afd834dc2d4d2a1ac0f8ce10d5b6b727f60776d/server/routes/food.js#L23

jspaine: Sorry I got it wrong above, you do have to be logged in to apply as a customer. So all you should have to do is remove the authorize check in this function.

PR #369, opened 2018-03-23
Add the displaying of password errors on the signup page

### LINKED ISSUE #364: "Password should be longer" should be displayed to the user while trying to register a new membership
https://pantry-for-good.herokuapp.com/users/signup
image

gomezvillegasdaniel:
closes #364

### DIFF
--- a/client/modules/users/components/SignUp.js
+++ b/client/modules/users/components/SignUp.js
@@ -28,9 +28,14 @@ class SignUp extends React.Component {
     this.props.clearFlags()
   }
 
-  onFieldChange = e => {
+  onFieldBlur = e => {
     const {name, value} = e.target
     this.setState({[name]: {value, touched: true}})
+  }
+
+  onFieldChange = e => {
+    const {name, value} = e.target
+    this.setState({[name]: {value, touched: this.state[name].touched ? true : false}})
   }
 
   onSubmit = e => {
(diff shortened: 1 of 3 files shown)


gomezvillegasdaniel:
@jspaine Hi, could you review this please?

jspaine:
Looks good, I think it's nice if password requirement errors are shown immediately though, before entering the confirm password. I guess this only shows after the confirm password field has been typed in? Can it be in an onBlur event on the password field?

gomezvillegasdaniel:
that's true, hands on it

gomezvillegasdaniel:
@jspaine how do you see it now?

client/modules/users/components/SignUp.js
```diff
@@ -28,11 +28,16 @@ class SignUp extends React.Component {
...
 
-  onFieldChange = e => {
+  setFieldAsTouched = e => {
     const {name, value} = e.target
     this.setState({[name]: {value, touched: true}})
   }
 
+  setFieldValue = e => {
+    const {name, value} = e.target
+    this.setState({[name]: {value}})
```
jspaine: this clears the touched state when the value is changed, so the errors disappear even if the new value is still invalid

client/modules/users/components/SignUp.js
```diff
@@ -138,10 +145,14 @@ class SignUp extends React.Component {
           {password.touched && passwordConfirm.touched && password.value !== passwordConfirm.value &&
             <div className="alert alert-danger">Passwords do not match</div>
           }
+          {password.touched && (password.value.length < 7 && passwordConfirm.value.length < 7) &&
```
jspaine: less important than the above, but I think it would be better if only one error was shown at a time - if the password requirements aren't met show the password must be ... error, otherwise show the passwords don't match error if appropriate

client/modules/users/components/SignUp.js
gomezvillegasdaniel: I did that because of the adding of the onBlur event you asked. The validation should be triggered on the onBlur event, not while typing. And, due to the validation is triggered while the element is touched, it displays or hides the error reactively. So I thought that it is better like this way.

jspaine: The usual way is the first validation is on blur, so you don't get an error while first filling out the field. Then validation happens on every change, so you can see when it's valid without having to leave the field.

client/modules/users/components/SignUp.js
gomezvillegasdaniel: done, thanks for the review

jspaine:
Thanks!

PR #374, opened 2018-04-30
Fixed disabled checkbox in packing page

### LINKED ISSUE #300: Error on Packing List page when trying to pack with no preferred foods

kristoferma:
Removed grey background from disabled checkbox and added not-allowed cursor.

screen shot 2018-04-30 at 4 20 31 pm

Closes #300

### DIFF
--- a/client/application.css
+++ b/client/application.css
@@ -1,11 +1,11 @@
 /* partially checkable checkboxes */
-input[type='checkbox'] {
+input[type="checkbox"] {
   position: absolute;
   opacity: 0;
   z-index: -1;
 }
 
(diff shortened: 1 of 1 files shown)


jspaine:
Looks good, thanks!

PR #378, opened 2018-05-26
Fixed issues with notifications of user status changes

### LINKED ISSUE #320: Changing customer status from Away to Available should not send an email
To replicate:
1. logon as the admin user
2. Go to the list of customers
3. Find an accepted customer and click the pencil icon to edit
4. On the edit page click the Away button in the status section
5. Then click the Available button to to set status to Available

When the Available button is clicked it sends the "Customer Accepted" email to the customer.  It should only send this email when the customer has been changed from Pending to Accepted.

#288 was a really similar issue so the fix for that may be similar as well.

kristoferma:
Changed status update notifications according to discussion on issue #320 

close #320

### DIFF
--- a/server/controllers/customer.js
+++ b/server/controllers/customer.js
@@ -50,9 +50,11 @@ export default {
     const oldCustomer = await Customer.findById(customer._id)
     const newCustomer = await customer.save()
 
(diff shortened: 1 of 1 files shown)


jspaine:
Thanks!

PR #391, opened 2018-07-05
add tests to client/modules/core/components

redbow-kimee:
Tested Footer.js and Title.js


jspaine (commented):
Looks good! Couple of mostly style points:

You don't need to `expect(t != null && t != undefined).to.eql(true)` or similar, the test will fail anyway otherwise.

I think `setProps` calls update itself.

And I'd try to keep all the setup to a minimum for a test, eg. the 'should display the org as the Title when there is one path element' doesn't use any of the function props so they can just be `() => {}`. For bigger tests especially it makes it a bit easier to see what's happening.

redbow-kimee:
Fixed

jspaine:
Thanks!

Issue #375, opened 2018-05-16
Notifications


jspaine:
- [ ] Remove notifications when they are viewed
- [ ] Tidy up notifications list
  - [ ] Don't show url
  - [ ] Format the date nicer
  - [ ] Does it need to be a table or would a list be better
- [ ] Other notifications
  - [ ] Volunteer created/updated/role changed
  - [ ] Donation receipt requested

If anyone can think of other notifications we might want, comment and I'll add them here.

ronaldblanco:
Hello @jspaine 
I will begin to work on this steps.

ronaldblanco:
Hello @jspaine 
Hey I am sorry I have been a little sort of time but I have been checking (to resolve the 1 point "Remove notifications when they are viewed") but I do not fine in the application other side were sometime like that it is working. Could you give me and AB of how to start this??

Thanks.

jspaine:
Hi, no problem. I think it's just to make it so when you click to view a notification you dispatch an action from the `deleteNotification` action creator before navigating.

The api call in `deleteNotification` should probably have `schema: null, responseSchema: arrayOfNotifications` though, because you're only sending a notification id and getting back the new list of notifications.

ronaldblanco:
Hey @jspaine take a look and let me know.

PR #392, opened 2018-07-12
fix/375#1 Remove notifications when they are viewed

ronaldblanco:
To fix 375#1 (Remove notifications when they are viewed)
The solution was tested locally and lint checked.
#375

### DIFF
--- a/client/modules/users/components/Notifications.js
+++ b/client/modules/users/components/Notifications.js
@@ -118,7 +118,6 @@ class NotificationsList extends Component {
                 <TableHeaderColumn dataField="index" width="70px" dataSort>#</TableHeaderColumn>
                 <TableHeaderColumn dataField="date" dataSort>Date</TableHeaderColumn>
                 <TableHeaderColumn dataField="message" dataSort>Message</TableHeaderColumn>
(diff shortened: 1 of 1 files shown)


ronaldblanco:
Hello @jspaine;
Take a look to this solution for the 375 #1 Issue and let me know.

Thanks.

jspaine:
Hi, it works but you should use the `Link` component or the whole page re-renders. Also clicked notifications in the dropdown in the top right corner don't get deleted.

ronaldblanco:
I do not understand, when I click to view anyway will navigated to the notification URL so you will refresh the page once you come-back.
I will be working in the dropdown.

jspaine:
Try it, the whole page reloads if you use a regular link, if you use react-router `Link` only the main section of the view changes. You can look at `window.history` in the console after clicking view notification - react router pushes the new url onto the history with a state object and triggers the view to update, when you go back it can just update the view again, with regular links there's no state in the history and the page just reloads at the current url.

You can also use `push` from react-router-redux and navigate in the `viewAndDeleteNotification` method (or change the name because at the moment all it does is delete the notification)

PR #393, opened 2018-07-14
fix security vulnerabilities as suggested by npm install using npm audit fix --force

spotlesscoder:


jspaine:
Thanks, but it seems there have been some [changes](https://fettblog.eu/gulp-4-parallel-and-series/) in gulp 4 and it doesn't build. Also webpack-dev-server 3 has a peer dependency on webpack 4 so I'm not sure it'll work without updating that too.

spotlesscoder:
Oh yes, indeed. Do you plan upgrading? 

Another thing: maybe it would be good to run "npm run dev" on Travis CI. This way, I would have noticed it myself. I was too confident that it would work ;)

PR #401, opened 2018-12-02
Added tests for EditUser.

JonSBoyce:
Made two changes to the file under test. First an export statement to make it testable in isolation of the react system and second to correct a typo.


jspaine:
Thanks, looks good!

PR #404, opened 2019-02-12
Add customer controller unit tests

tsukimi2:
Add customer controller unit test

### DIFF
--- a/server/controllers/customer.spec.js
+++ b/server/controllers/customer.spec.js
@@ -1,0 +1,327 @@
+const sinon = require('sinon')
+const sandbox = sinon.createSandbox()
+import {omit} from 'lodash'
+import customerCtrl from './customer'
+import Customer from '../models/customer'
+import User from '../models/user'
+import { customerStatus } from './../../common/constants'
+import Volunteer from '../models/volunteer'
+import Package from '../models/package'
+import mailer from '../lib/mail/mail-helpers'
+
+
+describe('Customer controller', function() {
+  describe('create', function() {
+    let req = {}
+    let res = {}
+
+    beforeEach(function() {
+      req = {
+        body: {
+          firstName: 'user',
+          lastName: 'test',
+          email: 'user@test.com',
+          __v: 0
+        },
+        user: {
+          id: 99999
+        }
+      }
+
+      res = {
+        json: sandbox.spy(),
+        status: sandbox.stub().returnsThis
+      }
+    })
+
+    afterEach(function() {
+      req = {}
+      res = {}
+      sandbox.restore()
+    })    
+
+    it('should return created customer object', async function() {      
+      const expectedResult = {    
+        fields: [],
+        firstName: req.body.firstName,
+        lastName: req.body.lastName,
+        _id: req.user.id
+      }
+
+      const customerSaveStub = sandbox.stub(Customer.prototype, 'save').returns(expectedResult)
+      const userFindStub = sandbox.stub(User, 'findOneAndUpdate')
+	
+      await customerCtrl.create(req, res)
+     
+      sinon.assert.calledOnce(userFindStub)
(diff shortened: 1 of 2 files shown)


server/controllers/customer.spec.js
```diff
@@ -0,0 +1,458 @@
+const sinon = require('sinon')
+const sandbox = sinon.createSandbox()
+import {omit} from 'lodash'
+const customerCtrl = require('./customer').default
```
jspaine: Why not just import this and sinon instead of requiring?

server/controllers/customer.spec.js
```diff
@@ -0,0 +1,458 @@
...
+          notifications: [ user.notifications[0] ],
+          id: user._id,
+          email: user.email,
+          provider: user.provider,
+        }
+      }
+
+      res = {
+        json: sandbox.spy(),
+        status: sandbox.stub().returns({ end: sandbox.spy() })
```
jspaine: Nothing calls res.status().end() so this isn't needed. Sinon spies/stubs have a `returnsThis` that can be useful:
```
res = {
  status: sandbox.stub().returnsThis,
  end: sandbox.spy()
}
```

server/controllers/customer.spec.js
```diff
@@ -0,0 +1,458 @@
...
+
+describe('Customer controller', function() {
+  describe('create', function() {
+    let req = {}
+    let res = {}
+    let user = null
+    let newCustomer = null
+    let userSession = null
+
+    before(async function() {
```
jspaine: I'd try to get rid of these, setting up the db shouldn't be needed for a unit test if everything is stubbed.

server/controllers/customer.spec.js
```diff
@@ -0,0 +1,458 @@
...
+    after(async function() {
+      await Questionnaire.find().remove()
+      await resetDb()
+      user = null
+    })
+
+    beforeEach(function() {
+      req = {
+        params: {},
+        body: {
```
jspaine, line 20: Most of `req.body`, `req.user` and `expectedResult` doesn't seem to be needed, I'd cut it down to only what the test is looking for.

server/controllers/customer.spec.js
```diff
@@ -0,0 +1,458 @@
...
+        household: [],
+        fields: [],
+        _id: 45633,
+        firstName: 'user',
+        lastName: 'test',
+        email: 'user@test.com'
+      }]
+
+      mockFindAll = {
+        sort: function() {
```
jspaine: Can simplify with `returnsThis` - `sort: sinon.stub().returnsThis`

jspaine (commented):
Thanks, it looks good. I'd try to cut down on the mock data though and get rid of the db setup (if everything's stubbed nothing should be calling the db)

server/controllers/customer.spec.js
```diff
@@ -0,0 +1,458 @@
...
+      }
+    })
+
+    afterEach(function() {
+      req = {}
+      res = {}
+      sandbox.restore()
+    })
+
+    it('should update customer and volunteer and return deleted customer object if the customer is associated with any packages', async function() {
```
jspaine: I'd nest these so it's not "it should do all the things", something like:
```
describe("when the customer has packages")
  it("should update the customer")
  it("should update the volunteer")
  it("should return the deleted customer")
describe("when the customer doesnt have packages")
  it("should respond 409")
```

tsukimi2:
Updated pull request with suggestions you made above.   However, after I pushed the amended commit up the repos, the Travis CI build failed, with the test for User Api Signup "1) does not allow signup if email address already has an account" failing.  When I run the test on my local repos, all the tests passed without problem.  Would you be kind enough to take a look at this problem? Thanks.

PR #408, opened 2019-03-12
Fix/common constants

### LINKED ISSUE #403: Use of common constants
While writing test code, I notice in the js code that there are a few places where some of the constants defined in common/constants.js could have been used.  Should I correct them as I notice them, or maybe even spend a little bit of time to search through the code and replace them?

For example:
Line 31-32 of server/models/customer.js
```js
const CustomerSchema = new Schema({
  ...
    status: {
    type: String,
    enum: ['Accepted', 'Rejected', 'Pending', 'Inactive'],
    default: 'Pending'
  },
...
```

the values 'Accepted', 'Rejected', 'Pending', 'Inactive' could be replaced with constants already defined in common/constants.js:
customerStatus.ACCEPTED, customerStatus.REJECTED, customerStatus.PENDING, customerStatus.INACTIVE.

Another example:
common/constants.js:

```js
export const ADMIN_ROLE = 'roles/admin'

export const clientRoles = {
  CUSTOMER: 'roles/customer',
  DONOR: 'roles/donor',
  VOLUNTEER: 'roles/volunteer'
}
```

where such values can be found to be used in a few places in the code but without referencing the common constants, for example:
Lines 28, 61 of server/controllers/customer.js:

```js
searchUserAndSetNotification('roles/admin', ...)
```

tsukimi2:
Replace literal values with corresponding constants defined in common constants.

closes #403

### DIFF
--- a/server/controllers/customer.js
+++ b/server/controllers/customer.js
@@ -25,13 +25,11 @@ export default {
     )
 
     // Sent Notification
-    searchUserAndSetNotification('roles/admin', {message:`Customer ${customer.fullName} was created!`, url: `/customers/${customer._id}`}, req.user._id)
+    searchUserAndSetNotification(ADMIN_ROLE, {message:`Customer ${customer.fullName} was created!`, url: `/customers/${customer._id}`}, req.user._id)
 
     const savedCustomer = await customer.save()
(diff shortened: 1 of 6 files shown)


jspaine:
Thanks!

PR #412, opened 2019-03-13
Add unit tests for read() of customer controller

tsukimi2:


tsukimi2:
@jspaine I tried to test out whether my code is part of the problem of triggering the test fail on a different part of the code and it seems that whenever I include the customerCtrl.read(req, res) (e.g. in line 92 of server/controllers/customer.spec.js), it will cause the test failure.  However, I still couldn't figure out how invoking the read function of the customer controller (which merely consists of the single line res.json(req.customer)) can actually trigger the test failure in a user API signup test.  I've been reading the code of both the controller controller and the user controller's signup but can't seem to see from the code how one can affect the other.  It'll be great if you could help me a bit on this issue.  Thanks.

PR #414, opened 2019-04-14
Add unit test files for controllers

tsukimi2:
Add unit test files for controllers


tsukimi2:
@jspaine Added unit tests for server controllers except for the user controller.

